PR #1, opened 2016-03-14
System tests initial

blaskovic:
Hi,

this PR is about new system tests for centos systemd CI testsuite. It contains just one test case for now and some code to handle it.

I will add another test cases in different PRs.


zonque:
I now hooked up these tests in slave-control.py, and the succeed. Thanks for sharing, looking forward to more such additions!

PR #6, opened 2016-09-05
Timedatectl

blaskovic:
New test for timedatectl


zonque:
I'm blindly merging this now. If it causes regressions, I will revert. Thanks!

blaskovic:
I've tested in on CentOS 7.2, but there could be some differences on your testing environment.

PR #7, opened 2016-09-05
tmpfiles

blaskovic:
This is failing for now but it should work in my opinion.

I think it is broken in upstream, so **please wait with merging** before https://github.com/systemd/systemd/issues/4097 is not resolved.

In case this is not an issue, I will adjust the testcase.


blaskovic:
This should work with the newest systemd. Can you merge it and try it out?

zonque:
Jup, sure. Let's see if there is any fall out :)

blaskovic:
Looks good https://ci.centos.org/job/systemd-pr-build/1087/consoleFull - ctrl+f: "1296288".
Great!

PR #8, opened 2016-10-30
Various fixes

evverx:
I tried to reproduce https://github.com/systemd/systemd/pull/4456#issuecomment-257065303
And that's not so obvious :-)
So, I fixed some small bugs and updated the `README`

@zonque , please  have a look

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,3 +66,20 @@ cd /home/systemd/systemd-centos-ci
 
 ./slave-control.py $ARGS
 ```
+
+# Manually running the tests
+```sh
+# bootstrap
+yum install -q -y git
+git clone https://github.com/systemd/systemd-centos-ci
+systemd-centos-ci/slave/bootstrap.sh pr:pr-number # for example pr:4456
+systemctl reboot
+
+# testsuite
+systemd-centos-ci/slave/testsuite.sh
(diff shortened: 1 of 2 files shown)


zonque:
LGTM. Will merge - let's see how that behaves in the CentOS CI farm :)

evverx:
@zonque , thanks!
I want to run the https://github.com/systemd/systemd/pull/4378
But I'm not sure how to properly install the `busybox`

* `yum install -q -y busybox` is not worknig
* `wget https://busybox.net/downloads/binaries/busybox-$(uname -p)` is the ugly way :-)

Maybe, upload to the https://copr-be.cloud.fedoraproject.org/results/lnykryn/systemd-centosci-environment/ ? @lnykryn ?

BTW:
I've found a bug. The `-U`-flag doesn't work on Centos (if `user_namespace.enable` is not set)
Details: https://gist.github.com/evverx/922acebc38f4a7c74eb5918dd11a6793

lnykryn:
Hmm, I wanted to build busybox package for el7 but the build is failing for some reason. I need to look at it more closer. But pull requests to https://github.com/lnykryn/systemd-centosci-environment are also welcomed :-)

PR #13, opened 2017-03-13
bootstrap: install python3 from software collections

msekletar:
I am currently working on https://github.com/systemd/systemd/pull/5354
and PR also introduces first python based test to TEST-* series. This
commit is prerequisite for running that test in Cent OS CI.

https://www.softwarecollections.org/en/scls/rhscl/rh-python35/


msekletar:
I figured out that python34 is also shipped in epel. Hence we don't need to install python3 from software collections.

PR #15, opened 2017-08-06
bootstrap: switch to meson

evverx:
This should be merged and enabled after https://github.com/systemd/systemd/pull/6544
is merged.


poettering:
not grokking too much of this, but looks superficially OK and the old code was definitely broken, hence merging!

@evverx thanks a ton for working on this, much appreciated

PR #17, opened 2018-10-17
[WIP] Latest fixes

mrc0mmand:
Attempt to revive the CentOS CI to its former glory. So far I fixed the compilation by creating a new copr repository with required dependencies ([here](https://copr.fedorainfracloud.org/coprs/mrc0mmand/systemd-centos-ci/)), replaced python 3.4 with python 3.6 (meson requirement), and several other lesser fixes.

The dependency repository mentioned above contains a `kernel-alt` package, which is the same package included in RHEL-ALT 7.6, just recompiled in EPEL buildroot (I'm not sure if there's a CentOS equivalent of RHEL-ALT). This is necessary as upstream systemd depends on features from newer kernels. I'm not sure if we want to use a third-party repo with newer kernel builds for CentOS, but that's for another discussion.

Following things still need to be done:

- [ ] Run all tests from the upstream testsuite (TEST-*) [slave/testsuite.sh]
- [x] Sync downstream tests (slave/tests)
- [x] Support beakerlib libraries
- [ ] Import missing downstream sanity tests
- [ ] Probably automate dependency rebuild in copr (right now it's more like a quick hack to make the CI working)
- [ ] Verify if it's indeed working in the CentOS CI environment (right now I'm using a local VM for quick tests)
- [ ] Re-enable the CentOS CI when everything works as expected

Any improvements and comments are more than welcome.


slave/tests/hostnamectl/Makefile
```diff
@@ -0,0 +1,65 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
keszybz: In the long term, I hope we can have just one canonical location so we don't need to "sync" things.

PR #20, opened 2018-10-23
Perform regression fuzz testing

evverx:
Reproducers for bugs uncovered with OSS-Fuzz are added regularly.
To make it a little bit more useful let's also compile the fuzzers
and pass those reproducers to them.

### DIFF
--- a/slave/bootstrap.sh
+++ b/slave/bootstrap.sh
@@ -8,7 +8,7 @@ curl "$COPR_REPO" -o "/etc/yum.repos.d/${COPR_REPO##*/}"
 yum -q -y install epel-release yum-utils
 yum-config-manager -q --enable epel
 yum -q -y update
-yum -q -y install systemd-ci-environment python-lxml python36 ninja-build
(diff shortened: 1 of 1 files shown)


evverx:
@mrc0mmand thank you for the review!

I think it would be great to add a hook to trigger the build when PRs are opened here. In theory, the master systemd branch will be always green :-) so it would be safe to run the test suite against it.

mrc0mmand:
That's definitely in my TODO list. I'm still waiting for credentials to the CentOS CI, so in the meanwhile I wanted to at least set up Travis/Semaphore. However, as they both use Ubuntu, I'd have to use Docker (or something similar) to get CentOS in there, which is yet another may-break-in-the-future layer...

evverx:
I wouldn't trust CentOS containers running on top of Ubuntu Trusty :-) so waiting for the "real" CentOS machine seems to be the best course of action.

And thanks again for bringing the CI back to life! At such a pace it's going to work again very soon!

PR #26, opened 2018-12-06
meson: fix nobody user and group

yuwata:
CentOS uses nfsnobody as UID/GID 65534.

Not tested. But I hope this fixes two warnings during build.


mrc0mmand:
Thanks for the patch, the warning is indeed fixed. There is a second warning during the fuzzer-compilation phase (`[1648/1733] Generating sanitize-address-fuzzers with a custom command.`), but I guess this would have to be fixed in the systemd configuration, right?

yuwata:
> I guess this would have to be fixed in the systemd configuration, right?

Yeah, right.

PR #27, opened 2018-12-08
Support artifact exporting

mrc0mmand:
This PR introduces artifact exporting in the CentOS CI infrastructure. 

The shenanigans with the environment and rsync password files are necessary to make the directory structure on the [artifact server](http://artifacts.ci.centos.org/systemd/) at least a little bit pretty, and to not expose sensitive information (like rsync password). Nevertheless, the directory structure is still far from pretty, at least in my opinion (`systemd/<branch|commit|pr>/<log-folder>/<log-file>.log>`) but let's see how it's going to work out.

### DIFF
--- a/agent-control.py
+++ b/agent-control.py
@@ -2,7 +2,7 @@
 
 # GPLv2 etc.
 
-import os, json, urllib, subprocess, sys, argparse, fcntl, time
+import os, json, urllib, subprocess, sys, argparse, fcntl, time, tempfile
 
 github_base = "https://github.com/systemd/"
 git_name = "systemd-centos-ci"
@@ -87,6 +87,12 @@ def reboot_host(host):
 
     reboot_count += 1
 
+def fetch_artifacts(host, remote_dir, local_dir):
+    exec_cmd(["/usr/bin/scp", "-r",
+            "-o UserKnownHostsFile=/dev/null",
+            "-o StrictHostKeyChecking=no",
+            "root@%s:%s" % (host, remote_dir), local_dir])
+
 def main():
     global debug
     global reboot_count
@@ -162,6 +168,9 @@ def main():
         else:
             branch = ''
 
+        # Create a temporary directory for artifacts
+        artifact_dir = tempfile.mkdtemp(prefix="artifacts_", dir=".")
+
         cmd = "yum install -y git && git clone %s%s.git" % (github_base, git_name)
         remote_exec(host, cmd)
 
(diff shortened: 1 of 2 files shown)


evverx:
@mrc0mmand basically I have no idea what you're doing here :-) so if you think it's ready to roll than there's no need to wait for me here. Just press the merge button.

I'm having trouble finding a link I can click to download the artifacts at https://ci.centos.org/job/systemd-ci-build/6/. I'm wondering if I'm missing something obvious.

mrc0mmand:
> I'm having trouble finding a link I can click to download the artifacts at https://ci.centos.org/job/systemd-ci-build/10/. I'm wondering if I'm missing something obvious.

Good point, the link is indeed missing there (right now it's visible only in the build console log), will look into that. The CentOS CI Jenkins is missing the Description Setter plugin, but from what I've found so far it can be easily done using a groovy script.

mrc0mmand:
I agree, journal dump would be definitely handy - added it to the exit handler. As for the job description - even a simple groovy script requires an authorization from the Jenkins admin. I'll open a ticket for the Jenkins [Description Setter plugin](https://wiki.jenkins.io/display/JENKINS/Description+Setter+Plugin) which should make things somewhat easier.

evverx:
If I recall correctly Fedora CI failed from time to time because there wasn't enough space for placing packages it built, which brings me to my last question. Is there any kind of rotation mechanism in place that would prevent CentOS CI from failing due to inability to upload artifacts to the servers where those artifacts are supposed to be stored?

mrc0mmand:
That's actually a good question, yet I don't know the answer. However, I started toying with an idea of using the Jenkins' internal artifact system instead of the external artifact storage. Will explore it a bit further if it's worth implementing, as it could (theoretically) save us some trouble regarding the artifact rotation.

mrc0mmand:
Sorry it took so long, I just couldn't find the right solution which both worked and was somehow acceptable (at least for me).

I ditched the rsync part completely, as it was unnecessarily complicated and introduced yet another point of possible failure. The artifact machinery provided by Jenkins is just enough for our needs, the only challenge is collecting of all required artifacts from the remote SUT. This is, of course, just a basic concept, I'm pretty sure there's plenty of other artifacts which should be exported to further help in future debugging.

Example build: https://ci.centos.org/job/systemd-ci-build/24/

#### Jekinks configuration (relevant for changes in this PR) for future reference:
By default the workspace is shared between build, but that causes mess during artifact exporting. This change crates a separate workspace for each build
- ZQM Event Publisher
    - Use custom workspace
      - Directory: `systemd-ci-build-$BUILD_NUMBER`

* Build Environment
    * Delete workspace before build starts

- Post-build Actions
    - Archive the artifacts
        - Files to archive: `artifacts_*/`
    - Delete workspace when build is done

PR #32, opened 2018-12-19
Do a basic sanity check before installing the new systemd

mrc0mmand:
In lights of recent issues with unbootable systems (e.g. https://github.com/systemd/systemd/pull/11178), it would make sense to do a basic sanity check if the new systemd is at least bootable before installing it to the SUT. This should provide precious debug information if something goes south and speed-up the feedback cycle a little bit.

Originally, I wanted to write a simple systemd-nspawn test case from scratch, but in the end using an existing case instead of reinventing the wheel makes more sense.

@evverx what do you think?

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -16,7 +16,7 @@ curl "$COPR_REPO" -o "/etc/yum.repos.d/${COPR_REPO##*/}"
 yum -q -y install epel-release yum-utils
 yum-config-manager -q --enable epel
 yum -q -y update
-yum -q -y install systemd-ci-environment python-lxml python36 ninja-build libasan
+yum -q -y install systemd-ci-environment python-lxml python36 ninja-build libasan net-tools strace nc busybox e2fsprogs quota dnsmasq qemu-kvm
 python3.6 -m ensurepip
 python3.6 -m pip install meson
 
@@ -66,6 +66,16 @@ CFLAGS='-g -O0 -ftrapv' meson build \
       -Dnobody-user=nfsnobody \
       -Dnobody-group=nfsnobody
 ninja-build -C build
+
+# Let's check if the new systemd at least boots before installing it
+# As the CentOS' systemd-nspawn version is too old, we have to use QEMU
+INITRD_PATH="/boot/initramfs-$(uname -r).img"
+KERNEL_PATH="/boot/vmlinuz-$(uname -r)"
+[ ! -f /usr/bin/qemu-kvm ] && ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-kvm
+make -C test/TEST-01-BASIC clean setup run clean-again TEST_NO_NSPAWN=1 INITRD=$INITRD_PATH KERNEL_BIN=$KERNEL_PATH KERNEL_APPEND=debug
+rm -f /usr/bin/qemu-kvm
+
+# Install the compiled systemd
 ninja-build -C build install
 
 cat >/usr/lib/systemd/system-shutdown/debug.sh <<_EOF_


agent/bootstrap.sh
```diff
@@ -66,6 +66,15 @@ CFLAGS='-g -O0 -ftrapv' meson build \
...
       -Dnobody-group=nfsnobody
 ninja-build -C build
+
+# Let's check if the new systemd at least boots before installing it
+# As the CentOS' systemd-nspawn version is too old, we have to use QEMU
+INITRD_PATH="/boot/initramfs-$(uname -r).img"
+KERNEL_PATH="/boot/vmlinuz-$(uname -r)"
+[ ! -f /usr/bin/qemu-kvm ] && ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-kvm
+make -C test/TEST-01-BASIC clean setup run clean-again TEST_NO_NSPAWN=1 INITRD=$INITRD_PATH KERNEL_BIN=$KERNEL_PATH KERNEL_APPEND="debug"
+rm -f /usr/bin/qemu-kvm
```
keszybz: This might remove the installed qem-kvm file. I'd just let it be. (Not that it matters too much either way.)

Looks reasonable.

mrc0mmand: This is actually on purpose and just temporary, as the integration test suite under KVM doesn't work completely (should be fixed soon by https://github.com/systemd/systemd/pull/11193).

agent/bootstrap.sh
```diff
@@ -66,6 +66,15 @@ CFLAGS='-g -O0 -ftrapv' meson build \
       -Dnobody-user=nfsnobody \
       -Dnobody-group=nfsnobody
 ninja-build -C build
+
+# Let's check if the new systemd at least boots before installing it
+# As the CentOS' systemd-nspawn version is too old, we have to use QEMU
+INITRD_PATH="/boot/initramfs-$(uname -r).img"
+KERNEL_PATH="/boot/vmlinuz-$(uname -r)"
+[ ! -f /usr/bin/qemu-kvm ] && ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-kvm
+make -C test/TEST-01-BASIC clean setup run clean-again TEST_NO_NSPAWN=1 INITRD=$INITRD_PATH KERNEL_BIN=$KERNEL_PATH KERNEL_APPEND="debug"
```
evverx: It would certainly have helped in https://github.com/systemd/systemd/pull/11178 and looks great overall. I'd turn SELinux off with `setenforce 0` before running the test, though.

evverx: Sorry. I apparently missed `TEST_NO_NSPAWN=1`. Running the test with QEMU should be fine with SELinux enforced.

agent/bootstrap.sh
```diff
@@ -66,6 +66,15 @@ CFLAGS='-g -O0 -ftrapv' meson build \
       -Dnobody-user=nfsnobody \
       -Dnobody-group=nfsnobody
 ninja-build -C build
+
+# Let's check if the new systemd at least boots before installing it
+# As the CentOS' systemd-nspawn version is too old, we have to use QEMU
```
evverx: In theory the test should be run with `systemd-nspawn` that has just been built. If it isn't, then something is wrong with the test. Could you open an issue?

agent/bootstrap.sh
```diff
@@ -66,6 +66,15 @@ CFLAGS='-g -O0 -ftrapv' meson build \
...
+INITRD_PATH="/boot/initramfs-$(uname -r).img"
```
evverx: We should probably regenerate the initrd file before running the test. On the other hand, making sure that the ancient version of systemd used on CentOS still can transition to a newly built one is kind of useful too :-)

agent/bootstrap.sh
mrc0mmand: Interesting, you're right. This was a quick guess without any further investigation, as the job was dying because of an unknown property:

```
Press ^] three times within 1s to kill container.
Failed to allocate scope: Cannot set property CollectMode, or unknown property.
Parent died too early
E: nspawn failed with exit code 1
```

Looking at the `test/test-functions` it looks the problem is somewhere else.

agent/bootstrap.sh
mrc0mmand: Right, I guess I should move the dracut part before the test, as it could affect the test results.

agent/bootstrap.sh
evverx: That looks suspiciously familiar. I hit something like that and even tried to "fix" it in https://github.com/systemd/systemd/pull/9750. As far as I understand, it's "not-a-bug" :-)

mrc0mmand: In that case I'd say it would be safer to just stick with QEMU, at least for this part of the CI.

agent/bootstrap.sh
evverx: For it to work properly `systemd` should be installed before `initrd` is regenerated but it's dangerous because the first `daemon-reload` can crash everything as happened with Ubuntu CI in https://github.com/systemd/systemd/pull/11178 where all the useful data was lost. I'm not sure what to do about it. I guess the question here is how unsafe it is not to upgrade initrd.

evverx: In theory, if it's possible to run the tests with `SKIP_INITRD=yes`, the whole `initrd` thing can be avoided. Could you try it to see how it goes? If it doesn't work I think it would be better to not regenerate it.

mrc0mmand: Unfortunately, the VM dies pretty painfully... I guess the multipath root fs is one of the culprits. Let's just not regenerate the initrd and see if it's going to work.

```
[    1.312595] List of all partitions:
[    1.314035] No filesystem could mount root, tried: 
[    1.316141] Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)
[    1.319690] CPU: 0 PID: 1 Comm: swapper/0 Not tainted 3.10.0-957.1.3.el7.x86_64 #1
[    1.322581] Hardware name: Red Hat KVM, BIOS 0.5.1 01/01/2011
[    1.324947] Call Trace:
[    1.326044]  [<ffffffff9d561e41>] dump_stack+0x19/0x1b
[    1.328059]  [<ffffffff9d55b550>] panic+0xe8/0x21f
[    1.329993]  [<ffffffff9db86761>] mount_block_root+0x291/0x2a0
[    1.332315]  [<ffffffff9db867c3>] mount_root+0x53/0x56
[    1.334499]  [<ffffffff9db86902>] prepare_namespace+0x13c/0x174
[    1.336818]  [<ffffffff9db863df>] kernel_init_freeable+0x1f8/0x21f
[    1.339424]  [<ffffffff9db85b1f>] ? initcall_blacklist+0xb0/0xb0
[    1.341981]  [<ffffffff9d54ff40>] ? rest_init+0x80/0x80
[    1.343821]  [<ffffffff9d54ff4e>] kernel_init+0xe/0x100
[    1.346068]  [<ffffffff9d574c24>] ret_from_fork_nospec_begin+0xe/0x21
[    1.348542]  [<ffffffff9d54ff40>] ? rest_init+0x80/0x80
[    1.351274] Kernel Offset: 0x1be00000 from 0xffffffff81000000 (relocation range: 0xffffffff80000000-0xffffffffbfffffff)```

evverx: It looks like `ext4` is shipped as a separate module. In theory, it should be possible to find out what filesystems are supported "natively" and pass something like `FSTYPE=btrfs` to `make` when it sets up the environment. Having said that, the current approach looks good to me too so we should probably stop here while it still works :-)

mrc0mmand: I'd leave it with the old initrd for now, so we have at least some debug output when something breaks. However, the `FSTYPE` shenanigans sound intriguing, will try to play with it in following days.

mrc0mmand:
Given the CI status I'll merge this PR to see how it's going to behave in the *real world*. Thanks a lot for all of your feedback!

agent/bootstrap.sh
mrc0mmand: Got a time to play around with using just a bootloader instead of an initrd for the QEMU, but without success, at least so far. No matter which filesystem I use (btrfs, ext*, xfs) the result is still the same. 

From what I've read, it should be enough to specify the root fs path on the kernel command line (`root=/dev/xxx`), but after several attempts using either full paths or UUIDs, I gave up. I'm not sure if I'm missing something fundamental or the initrd does something which is necessary to successfully boot the kernel.

PR #35, opened 2018-12-23
Add a copr repo mirror

mrc0mmand:
In light of recent copr outages, let's add a backup repo mirror,
so we can mitigate the outage impact as much as possible

(Technical background: simple reposync executed every hour)

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -19,6 +19,7 @@ set -e
 set -o pipefail
 
 COPR_REPO="https://copr.fedorainfracloud.org/coprs/mrc0mmand/systemd-centos-ci/repo/epel-7/mrc0mmand-systemd-centos-ci-epel-7.repo"
+COPR_REPO_PATH="/etc/yum.repos.d/${COPR_REPO##*/}"
 
 # Enable necessary repositories and install required packages
 #   - enable custom Copr repo with newer versions of certain packages (necessary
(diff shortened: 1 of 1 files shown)


mrc0mmand:
Just highlighting the most important lines of this patch:

```patch
+# Note: if a URL starts on a new line, it MUST begin with leading spaces,
+#       otherwise it will be ignored
+sed -i '/baseurl=/a\    https://rpm.sumsal.cz/mrc0mmand-systemd-centos-ci/' "$COPR_REPO_PATH"
+sed -i '/gpgkey=/a\    https://rpm.sumsal.cz/mrc0mmand-systemd-centos-ci/pubkey.gpg' "$COPR_REPO_PATH"
```

~~Not that it took me a while to find it out...~~

Anyway, after breaking the official copr URL (even with the GPG URL as well), yum seamlessly went on using the *mirror*, so it looks good to me.

cc @evverx

mrc0mmand:
Hm, thinking about this a little bit more, we could, in theory, *abuse* the CentOS CI infrastructure for the copr mirror. Either using the Jenkins' workspace, where the Jenkins would figure as both the storage and the job control, or using the CentOS CI artifact storage.

I'll make a PoC for the first variant, as I like the idea to have the dependencies in the same infrastructure as the CI itself.

mrc0mmand:
Well, I'd say technically it's possible (even though the Jetty directory listing is a little bit weird), however the CI slave is missing the `createrepo` package, which is necessary to generate repodata.

For now, I'd merge this PR, so we have at least some backup, and would get back to it once the `createrepo` *issue* is resolved.

PR #39, opened 2018-12-28
Visual improvements of test results

mrc0mmand:
Let's make fail investigation somewhat easier by adding several visual clues for the artifacts.

Namely:
- [ ] log file names should contain the test result (PASS/FAIL)
- [ ] as the artifact hierarchy may be somewhat confusing, let's generate a simple HTML page with a short description and artifact categorization (and by simple I mean really **simple**)
- [ ] ???
- [ ] profit!


mrc0mmand:
It seems to be working https://ci.centos.org/job/systemd-ci-build/104/artifact/index.html!

PR #47, opened 2019-01-05
bootstrap: update the initrd before the sanity check

mrc0mmand:
The original initrd provided by CentOS is pretty old (v219) and causes
several intermittent issues during the sanity check. Regenerating the
initrd however requires the new systemd to be installed, so let's
reorder the phases to ensure it.

@evverx I looked into the possibility of generating the initrd without installing the new systemd beforehand, and even though it is possible by writing a custom dracut config similar to the [default one](https://github.com/dracutdevs/dracut/blob/master/dracut.conf.d/fedora.conf.example), I'd say it's more fragile than just installing the new systemd and running the sanity check right after that.

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -92,6 +92,12 @@ esac
 echo -n "Checked out version "
 git describe
 
+# It's impossible to keep the local SELinux policy database up-to-date with
+# arbitrary pull request branches we're testing against.
+# Disable SELinux on the test hosts and avoid false positives.
+setenforce 0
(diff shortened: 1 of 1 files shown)


evverx:
I'd also turn selinux off with `setenforce 0` before systemd is installed so that it wouldn't crash if, for example, a rogue cron job tried to run `systemctl daemon-reexec` while the test was running. Other than that, the PR looks good to me. Thank you!

mrc0mmand:
Completely forgot about that, should be fixed now. Thanks!

PR #56, opened 2019-01-11
rhel8: create systemd-resolve user & group

mrc0mmand:
On RHEL 8 the default sysusers config file is missing the
systemd-resolve user, which is created during systemd-resolved RPM
installation. As we're compiling systemd from the source, create the
user and group manually


mrc0mmand:
Verified manually with RHEL8 systemd.

PR #60, opened 2019-01-21
meson: replace CFLAGS env variable with -Dc_args meson counterpart

mrc0mmand:
Thanks @keszybz for pointing this out.


keszybz:
Ack. IFAIK this is exactly the same, but certainly feels cleaner.

PR #66, opened 2019-02-06
Re-enable systemd networkd tests

mrc0mmand:
Thanks to Yu's and Susant's efforts all known failures of the `systemd-networkd-tets.py` should be fixed. For now I'll re-enable it only for the "vanilla" CI (i.e. the baremetal CentOS 7), and if it remains stable I'd like to enable it in the Vagrant builds as well.

### DIFF
--- a/agent/testsuite-rhel7.sh
+++ b/agent/testsuite-rhel7.sh
@@ -61,11 +61,10 @@ done
 ## Other integration tests ##
 TEST_LIST=(
     "test/test-exec-deserialization.py"
-#    "test/test-network/systemd-networkd-tests.py"
 )
 
 for t in "${TEST_LIST[@]}"; do
-    exectask "$t" "${t##*/}.log" "./$t"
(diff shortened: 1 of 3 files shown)


agent/testsuite-rhel7.sh
```diff
@@ -61,11 +61,10 @@ done
 ## Other integration tests ##
 TEST_LIST=(
     "test/test-exec-deserialization.py"
-#    "test/test-network/systemd-networkd-tests.py"
```
yuwata: Why the test is dropped, instead of uncomment the line?

mrc0mmand: That's just for RHEL7 - I'm not sure if it's worth the hassle trying to backport all the systemd-networkd tests (and related stuff) all the way back to RHEL7's systemd version (but backporting for RHEL8 is already in progress: https://github.com/lnykryn/systemd-rhel/pull/287)

PR #74, opened 2019-02-27
Make test timeouts * 3

keszybz:
I added a new test in https://github.com/systemd/systemd/pull/11841,
and it runs in ~8s on my machine, ~17s with address sanitizer. But the
test fails on centos-ci after a 30s timeout.


mrc0mmand:
LGTM, thanks!

PR #76, opened 2019-03-04
vagrant: bump QEMU_TIMEOUT to 2000s

mrc0mmand:
TEST-24-UNIT-TESTS started failing intermittently due
to newly introduced fuzzers. Bump the QEMU timeout to
reflect this change.

---

Should `fix` https://github.com/systemd/systemd/pull/11887#issuecomment-469296378

### DIFF
--- a/vagrant/vagrant-test.sh
+++ b/vagrant/vagrant-test.sh
@@ -36,7 +36,7 @@ for t in test/TEST-??-*; do
     ## Configure test environment
     # Set timeouts for QEMU and nspawn tests to kill them in case they get stuck
     # As we're not using KVM, bump the QEMU timeout quite a bit
-    export QEMU_TIMEOUT=1500
+    export QEMU_TIMEOUT=2000
     export NSPAWN_TIMEOUT=600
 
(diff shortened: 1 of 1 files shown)


mrc0mmand:
Relevant log part:
```
[TASK] test/TEST-24-UNIT-TESTS
Waiting for PID 20428 to finish

............................................................................................................................................................................................
PID 20428 finished with EC 0 in 1880s
[RESULT] PASS (log file: /build/vagrant-arch-testsuite.6bT/TEST-24-UNIT-TESTS_PASS.log)
```

evverx:
Looks like the wrapper that runs the tests in VMs and containers run them one by one. I think it should be possible to speed up the whole test by running the tests in parallel.

mrc0mmand:
> Looks like [the wrapper that runs the tests in VMs and containers](https://github.com/systemd/systemd/blob/master/test/TEST-24-UNIT-TESTS/testsuite.sh) runs them one by one. I think it should be possible to speed up the whole test by running the tests in parallel.

This idea actually crossed my mind as well, but I somehow dismissed it with *it definitely has it's reason why it's this way*. I'll definitely look into that, as this could significantly speed up all CentOS CI jobs.

PR #80, opened 2019-03-12
agent: remove stale state files

mrc0mmand:
TODO


mrc0mmand:
Should be fixed by https://github.com/systemd/systemd/pull/11981.

PR #85, opened 2019-03-18
vagrant: introduce openSUSE Tumbleweed

mrc0mmand:
I had some spare time on my hands and decided to play around with Vagrant, again. After a relatively short trial-and-error process I came up with an additional distro to our CI collection - openSUSE Tumbleweed. ~~I'm not even sure if we want to test systemd on it, I was just curious and never used openSUSE before~~.

@poettering @keszybz Do we want something like this in the CI? (I chose Tumbleweed over Leap because the Leap doesn't have all necessary deps for upstream systemd, much like CentOS).


mrc0mmand:
Closing in favor of https://github.com/systemd/systemd/issues/12040#issuecomment-475226472.

PR #87, opened 2019-03-21
Lower the number of concurrent jobs

mrc0mmand:
The preparation/booting of the QEMU/nspawn machines is
pretty I/O heavy operation, which started causing some severe
timing issues in some cases. Let's limit the job queue to
three concurrent tasks, to somewhat mitigate this.

---

This should fix the fail in systemd/systemd#12039, as the `machinectl import` command was timeouting due to a heavy disk I/O.


mrc0mmand:
After an another debug session this seems like a regression, as the test works as expected with the v241 release, will try to bisect it.

PR #91, opened 2019-04-07
common: initialize submodules when checking out a PR/commit

### LINKED ISSUE #90: CentOS CI doesn't seem to support submodules
It's more of a comment :-) and given that in general submodules aren't used in `systemd`, I'm not sure whether it should be fixed or not. Anyway, as https://github.com/systemd/systemd/pull/12151 shows, submodules are a convenient way to start experimenting with libraries that aren't available (because they aren't packaged, for example). Currently CentOS CI is bailing out complaining that "src/libbpf/meson.build:5:0: ERROR: File bpf_next/src/libbpf/src/bpf.c does not exist." (which is helpful in the sense that it shows that something unusual is happening) but, maybe, it would be better if PRs like that could be built and went through the testing machinery. Something like "git submodule update --init --recursive" after PRs are checked out should suffice I think. @mrc0mmand could you have a look?

mrc0mmand:
Let's support git submodules in the [systemd](https://github.com/systemd/systemd) repo.

Fixes #90

### DIFF
--- a/common/utils.sh
+++ b/common/utils.sh
@@ -24,6 +24,9 @@ git_checkout_pr() {
         esac
     ) || return 1
 
+    # Initialize git submodules, if any
+    git submodule update --init --recursive
+
     echo -n "[git_checkout_pr] Checked out version: "
     git describe
 }


mrc0mmand:
> Could you restart CentOS CI in [systemd/systemd#12151](https://github.com/systemd/systemd/pull/12151) after the PR is merged? Just to be sure :-)

Will do!

PR #92, opened 2019-04-10
common: fall back to head ref when merge ref doesn't exist

mrc0mmand:
Let's fix cases when a **merge** ref doesn't exist, i.e. for already merged pull requests and draft pull requests. Fall back to the (hopefully) always present **head** ref in such cases.

Should fix cases like systemd/systemd#12265.

### DIFF
--- a/common/utils.sh
+++ b/common/utils.sh
@@ -13,7 +13,14 @@ git_checkout_pr() {
     (set -e
         case $1 in
             pr:*)
-                git fetch -fu origin "refs/pull/${1#pr:}/merge:pr"
+                # Draft and already merged pull requests don't have the 'merge'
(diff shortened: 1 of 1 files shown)


mrc0mmand:
@evverx as for the rebase of the PR branch - there is apparently no way to get the *base* branch of the PR via the *standard means*. Should we worry about someone opening a draft PR against a non-master branch or just go with simple `git rebase master pr`?

mrc0mmand:
Well, we could do something like:

```
PR=11854; curl https://api.github.com/repos/systemd/systemd/pulls/$PR | jq .base.ref
```

but I'm not sure if it's not overengineering.

PR #93, opened 2019-04-10
agent: stress test TEST-16-EXTEND-TIMEOUT

mrc0mmand:
Let's see how stable the test actually is.


mrc0mmand:
See https://github.com/systemd/systemd/pull/12299

PR #96, opened 2019-04-18
Call ninja test with --print-errorlogs for easier troubleshooting

mrc0mmand:
Let's show the first couple of errors from `ninja test` in the console log,
so one doesn't have to dig through the enormous meson testlog.txt file.


mrc0mmand:
The failing compilation is being fix in https://github.com/lnykryn/systemd-rhel/pull/328

PR #99, opened 2019-05-03
Run tests on Arch Linux under ASan & UBSan

mrc0mmand:
Let's experiment with ASan & UBSan on Arch. If it's feasible, make it part of our regular testing workflow, as Travis CI doesn't cover tests, which require a full-fledged VM.

### DIFF
--- a/vagrant/README.md
+++ b/vagrant/README.md
@@ -1,0 +1,55 @@
+# systemd CentOS CI under Vagrant
+
+As the *standard* CentOS CI tests systemd on CentOS 7, which is quite old (and allows
+us to test compatibility with older kernels), we need some way to do the opposite - i.e.
+test systemd with the newest packages. To achieve this, we use
+[Vagrant](https://www.vagrantup.com/) along with [vagrant-libvirt](https://github.com/vagrant-libvirt/vagrant-libvirt)
+plugin to *easily* deploy libvirt VMs (with KVM acceleration) and run tests
+on [Arch Linux](https://www.archlinux.org/).
+
+The, so called, *pipeline* consists of following steps:
+
+1. `vagrant-ci-wrapper.sh`
+
+    This script is executed by our [Jenkins](https://jenkins.io/) job, and basically
+    prepares a local copy of the systemd [repository](https://github.com/systemd/systemd),
+    checks out a correct branch and executes the build script (see below) for
+    each distribution we have configured for Vagrant.
+
+    As Vagrant is not present in CentOS repositories, `vagrant-ci-wrapper.sh` also
+    makes use of `vagrant-setup.sh`, which downloads & installs the [upstream RPM](https://www.vagrantup.com/downloads.html)
+    provided by Vagrant, install libvirt & other dependencies, and compiles & installs
+    the vagrant-libvirt plugin.
+
+2. `vagrant-build.sh`
+
+    The Vagrant VM management is done by `vagrant-build.sh`, which locates the correct
+    Vagrantfile for provided distro tag, configures the environment appropriately,
+    spins up the VM, executes a test script according to the environment, and cleans up
+    the VM afterwards.
+
+    The compilation itself (along with installation of build & test dependencies)
+    is done during the VM provision phase using a provision script in a respective
+    Vagrantfile, as these steps are distro-specific and the CI scripts should be
+    (for the most part) distro-agnostic.
+
+    So far there are two tests scripts, for testing with and without sanitizers:
+
+    * `vagrant-test.sh`
+
+        Test systemd without sanitizers - do a reboot after installation and run unit tests,
+        fuzzers, and integration tests
+
+    * `vagrant-test-sanitizers.sh`
+
+        Test systemd with sanitizers ([Address Sanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) and
+        [Undefined Behavior Sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)).
+        As running under sanitizers has a huge performance impact, we skip machine reboot,
+        as this usually kills the machine or causes boot to timeout, and run only unit tests
+
+
+As we run the CI usually dozens of times per day, we'd have to update the packages
+in images each time (not counting the installation of build & test dependencies).
+To avoid this, and boost the runtime of the CI, we periodically execute `vagrant-make-cache.sh`
+which does these steps once in a few days, allowing us to reuse already up-to-date images
+in our CI runs without wasting time.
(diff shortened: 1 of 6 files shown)


mrc0mmand:
@evverx I had to bump the `--timeout-multiplier` all the way up to 15, due to hashmap tests (and a few others). However, thanks to #101, the runtime is still < 25m, which sounds reasonable. This will run (sequentially) in in the same job as the regular Vagrant Arch job, but again, thanks to #101, the runtime of the regular one went down from 55m to ~30m, so we should be still able to deliver results in an hour or so.

The two remaining failures are tracked in https://github.com/systemd/systemd/issues/12484 and https://github.com/systemd/systemd/pull/12488. Apart from that, the ASan/UBSan Vagrant job looks ready to go.

mrc0mmand:
Small edit: I had to disable certain parts of two tests, as `--timeout-multiplier=15` was still not enough for them:

https://github.com/systemd/systemd-centos-ci/blob/83978d9b56cfb995efb7177a3e5a3628083e619c/vagrant/vagrant-test-sanitizers.sh#L20-L28

evverx:
Regarding `test-capability`, it seems to be failing because `NOBODY_USER_NAME` doesn't exist there, which most likely means that `-Dnobody-user=nfsnobody` and `-Dnobody-group=nfsnobody` shouldn't be passed to `meson` in the first place. To judge from the build log, `meson` is complaining that "The local user with the UID 65534 does not match the configured user name "nfsnobody" of the nobody user (its name is nobody)." so it looks like both those options should be left out so that the test wouldn't be skipped.

mrc0mmand:
@evverx good catch, thank you! However, this also found a bug in the test itself, as described in https://github.com/systemd/systemd/pull/12488.

evverx:
@mrc0mmand I saw that PR :-) It helped me to find out that NOBODY_USER_NAME didn't exist there.

I've been thinking about what to do about the fuzzers (and `-Dslow-tests=true` in general). I think that at this point I'd simply leave it out to avoid the flakiness that will follow for sure, because, well, the fuzzers aren't quite ready to be run in environments where `ASAN_OPTIONS` is passed explicitly. The other "slow" tests have never been run under ASan+UBsan (on a regular basis at least) so they should probably be prepared for that as well.

By the way, I'm wondering why `-Ddbuspolicydir=/etc/dbus-1/system.d` doesn't break anything. I thought Arch ships with the latest version (more or less) of dbus where the kludge isn't necessary.

evverx:
I took a look at the package at https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/systemd and it seems that `-Ddbuspolicydir=/usr/share/dbus-1/system.d` is used there. Maybe we should use it too (or rely on the default).

mrc0mmand:
> @mrc0mmand I saw that PR :-) It helped me to find out that NOBODY_USER_NAME didn't exist there.
> 
> I've been thinking about what to do about the fuzzers (and `-Dslow-tests=true` in general). I think that at this point I'd simply leave it out to avoid the flakiness that will follow for sure, because, well, the fuzzers aren't quite ready to be run in environments where `ASAN_OPTIONS` is passed explicitly. The other "slow" tests have never been run under ASan+UBsan (on a regular basis at least) so they should probably be prepared for that as well.

Well, the issue with `-Dslow-tests=true` is that most of these tests are relatively fast, but there's a few *extra slow* tests which cause issues. So far, the only one, which kept failing all the time (`test-hashmap`), should be skipped (or at least the slowest part of the test).

I can't speak for the fuzzers, as my knowledge there is still pretty limited, but so far they've been behaving nicely (except for the two ones mentioned in this thread). 

I'd, probably, leave the slow tests enabled and watch them for few days how stable they are. 

> 
> By the way, I'm wondering why `-Ddbuspolicydir=/etc/dbus-1/system.d` doesn't break anything. I thought Arch ships with the latest version (more or less) of dbus where the kludge isn't necessary.

I don't even remember how it got there in the first place. I guess I just simply copied over compilation options from CentOS without much thinking about it.

> I took a look at the package at https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/systemd and it seems that `-Ddbuspolicydir=/usr/share/dbus-1/system.d` is used there. Maybe we should use it too (or rely on the default).

Will do, thanks for looking it up!

evverx:
> I'd, probably, leave the slow tests enabled and watch them for few days how stable they are.

Even if they're stable on CentOS CI, it won't help to simply turn the slow tests on on Travis CI. I mean, although I don't have problem with adding the workarounds to make the tests pass, those issues should be fixed on the systemd side.

mrc0mmand:
> Even if they're stable on CentOS CI, it won't help to simply turn the slow tests on on Travis CI. I mean, although I don't have problem with adding the workarounds to make the tests pass, those issues should be fixed on the systemd side.

Just to make things clear - should I disable all slow tests in CentOS CI (under sanitizers) or leave them enabled & report failures to systemd? (I haven't slept much and even coffee doesn't work anymore, so please just bear with me :-)

evverx:
I think for the time being it would make sense to turn the slow tests off (by removing `-Dslow-tests=true`) when systemd is built with ASan+UBSan because, to judge from Travis CI, we can be more or less sure that it kind of works. Plus we won't have to worry about https://github.com/systemd/systemd/issues/12484 :-)

Regarding reporting issues, I'm not sure. On one hand, it would be helpful to keep the list of known issues somewhere in case anyone is willing to fix them. On the other hand, usually nobody is interested in fixing them so it's kind of futile I think.

mrc0mmand:
Gotcha, that makes sense, at least we can drop the kludges for slow tests & uncooperative fuzzers.

evverx:
To be fair, the fuzzers kind of works as expected. It's just that when they were integrated into the test suite we didn't think of the testcases that triggers timeouts in the sense that currently all the fuzzers (compared to the ones triggering memory-leaks, buffer-overflows and so on). Franky, I still have no idea how to handle this properly :-)

evverx:
Not that it matters but now that the slow tests are turned off we could probably lower the timeout multiplier to, say, 3 so that the tests would fail faster should something goes wrong.

mrc0mmand:
> Not that it matters but now that the slow tests are turned off we could probably lower the timeout multiplier to, say, 3 so that the tests would fail faster should something goes wrong.

Good point, completely forgot about that, thanks!

mrc0mmand:
Looks like everything's ready to go. Thanks a lot for the help, @evverx!

evverx:
Thank you! I hope https://github.com/systemd/systemd/pull/12419 was the last PR where anyone had to run the unit tests under Valgrind/ASan manually now that CentOS CI and Travis CI cover both containers and VMs. In theory it should be safe to run TEST-01-BASIC under ASan+UBSan as we do in https://github.com/systemd/systemd/blob/master/travis-ci/managers/xenial.sh but with access to newer versions of gcc/ASan/UBSan. I'd wait for the unit tests to settle though :-)

Issue #100, opened 2019-05-03
Introduce caching for Vagrant images


mrc0mmand:
The initial Vagrant setup (system upgrade, installation of dependencies, ...) takes a significant amount of time. Consider creating another CentOS CI Jenkins job, which would periodically (e.g. once per week) download a new version of an image, install necessary dependencies. and cache it on the CentOS CI artifact server.

mrc0mmand:
Also, making a custom Vagrant package for CentOS 7 would make sense, so the CI doesn't have to rebuild the libvirt extension every time.

mrc0mmand:
#101 introduced a box/image cache, which is regenerated every 3 days by a [Jenkins job](https://ci.centos.org/job/systemd-centos-ci-vagrant-make-cache/).

The custom Vagrant package task seems to be more complex due to lack of dependencies, which would make everything more complex and generally not worth it.

PR #101, opened 2019-05-04
vagrant: introduce Vagrant image/box cache

mrc0mmand:
Let's update, rebuild, and cache Vagrant images, which we use in systemd CentOS CI.

This should have a pretty substantial performance impact, as the *setup* phase, which involves updating the base system & installing test and build dependencies usually takes ~10 minutes.

### DIFF
--- a/agent-control.py
+++ b/agent-control.py
@@ -277,6 +277,33 @@ class AgentControl(object):
         # Give the node time to finish booting process
         time.sleep(30)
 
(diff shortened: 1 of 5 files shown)


mrc0mmand:
Due to workarounds in `vagrant/vagrant-make-cache.sh` there should be a followup soon, but until then it should significantly improve the duration of the Vagrant CI job, fix half of #100, and allow us to introduce ASan and UBSan in #99.

PR #83, opened 2019-03-13
vagrant: enable systemd-networkd tests

mrc0mmand:
Let's try to enable `systemd-networkd-tests.py` under Vagrant (and more specifically under Arch), so we can test the new network-related features as well. 

After recent CI tweaks the test should be almost as fast as its baremetal counterpart, so hopefully it won't slow down the entire Vagrant job too much.

PR #104, opened 2019-05-07
vagrant: run TEST-01-BASIC under sanitizers

### LINKED ISSUE #103: Run TEST-01-BASIC under sanitizers
Run TEST-01-BASIC under ASan & UBSan as well, as this should be relatively safe and should test the systemd itself (and its components) under sanitizers.

See https://github.com/systemd/systemd-centos-ci/pull/99#issuecomment-489785233

mrc0mmand:
Let's run TEST-01-BASIC under sanitizers to cover the systemd daemon as a whole plus its subcomponents (like systemd-udevd).

Fixes #103

### DIFF
--- a/vagrant/vagrant-test-sanitizers.sh
+++ b/vagrant/vagrant-test-sanitizers.sh
@@ -18,7 +18,30 @@ export ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check
 export UBSAN_OPTIONS=print_stacktrace=1:print_summary=1:halt_on_error=1
 
 # Run the internal unit tests (make check)
-exectask "ninja-test" "meson test -C build --print-errorlogs --timeout-multiplier=3"
+exectask "ninja-test_sanitizers" "meson test -C build --print-errorlogs --timeout-multiplier=3"
+
+## Run TEST-01-BASIC under test sanitizers
+# Set timeouts for QEMU and nspawn tests to kill them in case they get stuck
+# As we're not using KVM, bump the QEMU timeout quite a bit
+export QEMU_TIMEOUT=600
+export NSPAWN_TIMEOUT=600
+# Set QEMU_SMP to speed things up
+export QEMU_SMP=$(nproc)
+# Arch Linux requires booting with initrd, as all commonly used filesystems
+# are compiled in as modules
+export SKIP_INITRD=no
+
+# 1) Run it under systemd-nspawn
+rm -fr /var/tmp/systemd-test*
+exectask "TEST-01-BASIC_sanitizers-nspawn" "make -C test/TEST-01-BASIC clean setup run clean-again TEST_NO_QEMU=1"
+# Each integration test dumps the system journal when something breaks
+[[ -d /var/tmp/systemd-test*/journal ]] && rsync -aq /var/tmp/systemd-test*/journal "$LOGDIR/TEST-01-BASIC_sanitizers-nspawn/"
+
+# 2) Run it under QEMU
+rm -fr /var/tmp/systemd-test*
+exectask "TEST-01-BASIC_sanitizers-qemu" "make -C test/TEST-01-BASIC clean setup run clean-again TEST_NO_NSPAWN=1"
+# Each integration test dumps the system journal when something breaks
+[[ -d /var/tmp/systemd-test*/journal ]] && rsync -aq /var/tmp/systemd-test*/journal "$LOGDIR/TEST-01-BASIC_sanitizers-qemu/"
 
 # Summary
 echo
--- a/agent/bootstrap-rhel8.sh
+++ b/agent/bootstrap-rhel8.sh
@@ -139,6 +139,7 @@ fi
             -Dtests=unsafe
             -Dinstall-tests=true
             -Dc_args='-g -O0 -ftrapv'
+            --werror
     )
     meson build "${CONFIGURE_OPTS[@]}"
     ninja-build -C build
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -89,6 +89,7 @@ systemctl disable firewalld
 #   - install-tests=true: necessary for test/TEST-24-UNIT-TESTS
 (
     meson build -Dc_args='-g -O0 -ftrapv' \
+                --werror \
                 -Dslow-tests=true \
                 -Dtests=unsafe \
                 -Dinstall-tests=true \
--- a/vagrant/Vagrantfiles/Vagrantfile_arch
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch
@@ -81,7 +81,9 @@ Vagrant.configure("2") do |config|
 
     rm -fr build
     # Build phase
-    CFLAGS='-g -O0 -ftrapv' meson build \
+    meson build \
+          --werror \
+          -Dc_args='-g -O0 -ftrapv' \
           -Dslow-tests=true \
           -Dtests=unsafe \
           -Dinstall-tests=true \
--- a/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers
@@ -83,7 +83,9 @@ Vagrant.configure("2") do |config|
     # Build phase
     # Compile systemd with the Address Sanitizer (ASan) and Undefined Behavior
     # Sanitizer (UBSan)
-    CFLAGS='-g -O0 -ftrapv' meson build \
+    meson build \
+          --werror \
+          -Dc_args='-g -O0 -ftrapv' \
           -Dtests=unsafe \
           -Dinstall-tests=true \
           -Ddbuspolicydir=/usr/share/dbus-1/system.d \


mrc0mmand:
@evverx I guess this might need some tweaking of the ASan wrapper in the systemd's `test-functions` file, because apparently Arch's kernel has all* filesystems compiled in as modules, thus boot without initrd ends up with a kernel panic:

```
[    3.449226] List of all partitions:
[    3.449471] No filesystem could mount root, tried: 
[    3.449506] 
[    3.449788] Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)
[    3.449788] CPU: 3 PID: 1 Comm: swapper/0 Not tainted 5.0.12-arch2-1-ARCH #1
[    3.449788] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.12.0-20181126_142135-anatol 04/01/2014
[    3.449788] Call Trace:
[    3.449788]  dump_stack+0x5c/0x80
[    3.449788]  panic+0x101/0x2a7
[    3.449788]  mount_block_root+0x2c2/0x2e6
[    3.449788]  ? do_early_param+0x8e/0x8e
[    3.449788]  prepare_namespace+0x130/0x166
[    3.449788]  kernel_init_freeable+0x2a5/0x2b4
[    3.449788]  ? rest_init+0xbf/0xbf
[    3.449788]  kernel_init+0xa/0x101
[    3.449788]  ret_from_fork+0x35/0x40
[    3.449788] Kernel Offset: 0x3b400000 from 0xffffffff81000000 (relocation range: 0xffffffff80000000-0xffffffffbfffffff)
[    3.449788] ---[ end Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0) ]---
```

* I tried ext2, ext3, ext4, and xfs

The VM does boot when I manually override `SKIP_INITRD=yes` [here](https://github.com/systemd/systemd/blob/master/test/test-functions#L51-L57):
```bash
if [[ "$IS_BUILT_WITH_ASAN" = "yes" ]]; then
    STRIP_BINARIES=no
    SKIP_INITRD=yes
    PATH_TO_INIT=$ROOTLIBDIR/systemd-under-asan
    QEMU_MEM="1536M"
    QEMU_SMP=4
fi
```

and by override I mean `sed -i '/SKIP_INITRD=yes/d' test/test-functions`. However, this skips the custom ASan initscript.

I'm not sure, what are the options here, but one of them could be moving to `SKIP_INITRD="${SKIP_INITRD:-yes}"`, so it can be overridden from the outside, and then customizing the default Arch's initrd, via `mkinitcpio`, by replacing the default initscript (which should be a symlink to the systemd binary) by our custom ASan wrapper.

We could run it only under nspawn, but that would skip udevd and other components...

evverx:
I see. Indeed, the asan wrapper should be changed to get it to work with initrd smoothly.

> We could run it only under nspawn, but that would skip udevd and other components...

I think it's a good idea in the sense that at least some part of the code is going to be covered (which is better than nothing) until the wrapper is fixed.

evverx:
> by override I mean sed -i '/SKIP_INITRD=yes/d' test/test-functions. However, this skips the custom ASan initscript.

@mrc0mmand I can't seem to find the logs I could look at to figure out why exactly that script was skipped and what happened instead when `SKIP_INITRD=yes` was used. Could you point me in the right direction?

evverx:
> I suspect that's because the ASan wrapper remounts the / as rw, which should be done by the systemd-remount-fs.service

The service shouldn't fail in this case and as far as I know it has never failed on Ubuntu Xenial on Travis CI where the same script remounting "/" is run. It could be that libasan somehow interacts weirdly with `mount` on Arch. Could you try to spin up a VM on Arch with `make clean setup run KERNEL_APPEND=systemd.unit=rescue.target` and then run something like
```sh
mount -o remount,rw /
echo $?

LD_PRELOAD= mount -o remount,rw /
echo $?
```

mrc0mmand:
@evverx ah, this was actually a misinformation on my side, as the script is not skipped (when booting with an initrd), see https://ci.centos.org/job/systemd-ci-build/434/artifact//artifacts_LVjBb8/vagrant-logs.Oda/console-arch-sanitizers.log - sorry for that

The only apparent issue (so far) seems to be failing `systemd-remount-fs.service`:

```
$ curl -q https://ci.centos.org/job/systemd-ci-build/434/artifact//artifacts_LVjBb8/vagrant-logs.Oda/console-arch-sanitizers.log | less -r

...
[FAILED] Failed to start Remount Root and Kernel File Systems.
See 'systemctl status systemd-remount-fs.service' for details.
...
systemd-remount-fs.service loaded failed failed Remount Root and Kernel File Systems
-rw-r-----+ 1 root systemd-journal 8388608 May  7 06:56 /var/tmp/systemd-test.kQ8Dop/journal/fa85b6d8047d4cd99496f560ef36bf18/system.journal
make: Leaving directory '/build/test/TEST-01-BASIC'
make: *** [Makefile:4: run] Error 1
```

mrc0mmand:
```
You are in rescue mode. After logging in, type "journalctl -xb" to view
system logs, "systemctl reboot" to reboot, "systemctl default" or "exit"
to boot into default mode.
Press Enter for maintenance
(or press Control-D to continue): 
bash-5.0# mount -o remount,rw /
bash-5.0# echo $?
1
bash-5.0# LD_PRELOAD= mount -o remount,rw /
ho $?bash-5.0# echo $?
0
```

It looks like libasan indeed interact with `mount` in some weird & unexpected way:
```
bash-5.0# mount -o remount,rw -v /
mount: /dev/sda1 mounted on /.
bash-5.0# echo $?
1
```

```
bash-5.0# strace -s 512 mount -o remount,rw -v /
...
ioctl(1, TCGETS, {B9600 opost isig icanon echo ...}) = 0
write(1, "mount: /dev/sda1 mounted on /.\n", 31mount: /dev/sda1 mounted on /.
) = 31
close(1)                                = 0
close(2)                                = 0
gettid()                                = 355
...
getpid()                                = 355
write(2, "==355==LeakSanitizer has encountered a fatal error.\n", 52) = -1 EBADF (Bad file descriptor)
write(2, "ReportFile::Write() can't output requested buffer!\n", 51) = -1 EBADF (Bad file descriptor)
exit_group(1)                           = ?
+++ exited with 1 +++

```

mrc0mmand:
I'm not sure how to properly debug this, as something is sending SIGKILL or SIGSTOP to `mount`:

With `verbosity=1`:
```
bash-5.0# export LSAN_OPTIONS=verbosity=1:log_threads=1:print_summary=1:detect_leaks=1:leak_check_at_exit=1:handle_segv=0:handle_sigill=0:handle_sigfpe=0:handle_sigbus=0
bash-5.0# mount -o remount,rw -v /
==433==AddressSanitizer: failed to intercept '__isoc99_printf'
==433==AddressSanitizer: failed to intercept '__isoc99_sprintf'
==433==AddressSanitizer: failed to intercept '__isoc99_snprintf'
==433==AddressSanitizer: failed to intercept '__isoc99_fprintf'
==433==AddressSanitizer: failed to intercept '__isoc99_vprintf'
==433==AddressSanitizer: failed to intercept '__isoc99_vsprintf'
==433==AddressSanitizer: failed to intercept '__isoc99_vsnprintf'
==433==AddressSanitizer: failed to intercept '__isoc99_vfprintf'
==433==AddressSanitizer: failed to intercept 'xdr_quad_t'
==433==AddressSanitizer: failed to intercept 'xdr_u_quad_t'
==433==AddressSanitizer: libc interceptors initialized
|| `[0x10007fff8000, 0x7fffffffffff]` || HighMem    ||
|| `[0x02008fff7000, 0x10007fff7fff]` || HighShadow ||
|| `[0x00008fff7000, 0x02008fff6fff]` || ShadowGap  ||
|| `[0x00007fff8000, 0x00008fff6fff]` || LowShadow  ||
|| `[0x000000000000, 0x00007fff7fff]` || LowMem     ||
MemToShadow(shadow): 0x00008fff7000 0x000091ff6dff 0x004091ff6e00 0x02008fff6fff
redzone=16
max_redzone=2048
quarantine_size_mb=256M
thread_local_quarantine_size_kb=1024K
malloc_context_size=30
SHADOW_SCALE: 3
SHADOW_GRANULARITY: 8
SHADOW_OFFSET: 0x7fff8000
==433==T0: stack [0x7fff49b06000,0x7fff4a306000) size 0x800000; local=0x7fff4a304da4
==433==AddressSanitizer Init done
mount: /dev/sda1 mounted on /.
Killed
bash-5.0# echo $?
137
```

With `verbosity=2` and above:
```
bash-5.0# export LSAN_OPTIONS=verbosity=2:log_threads=1:print_summary=1:detect_leaks=1:leak_check_at_exit=1:handle_segv=0:handle_sigill=0:handle_sigfpe=0:handle_sigbus=0
bash-5.0# mount -o remount,rw -v /
==436==AddressSanitizer: failed to intercept '__isoc99_printf'
==436==AddressSanitizer: failed to intercept '__isoc99_sprintf'
==436==AddressSanitizer: failed to intercept '__isoc99_snprintf'
==436==AddressSanitizer: failed to intercept '__isoc99_fprintf'
==436==AddressSanitizer: failed to intercept '__isoc99_vprintf'
==436==AddressSanitizer: failed to intercept '__isoc99_vsprintf'
==436==AddressSanitizer: failed to intercept '__isoc99_vsnprintf'
==436==AddressSanitizer: failed to intercept '__isoc99_vfprintf'
==436==AddressSanitizer: failed to intercept 'xdr_quad_t'
==436==AddressSanitizer: failed to intercept 'xdr_u_quad_t'
==436==AddressSanitizer: libc interceptors initialized
|| `[0x10007fff8000, 0x7fffffffffff]` || HighMem    ||
|| `[0x02008fff7000, 0x10007fff7fff]` || HighShadow ||
|| `[0x00008fff7000, 0x02008fff6fff]` || ShadowGap  ||
|| `[0x00007fff8000, 0x00008fff6fff]` || LowShadow  ||
|| `[0x000000000000, 0x00007fff7fff]` || LowMem     ||
MemToShadow(shadow): 0x00008fff7000 0x000091ff6dff 0x004091ff6e00 0x02008fff6fff
redzone=16
max_redzone=2048
quarantine_size_mb=256M
thread_local_quarantine_size_kb=1024K
malloc_context_size=30
SHADOW_SCALE: 3
SHADOW_GRANULARITY: 8
SHADOW_OFFSET: 0x7fff8000
==436==SetCurrentThread: 0x7f057f786000 for thread 0x7f057faed800
==436==T0: stack [0x7ffe0c353000,0x7ffe0cb53000) size 0x800000; local=0x7ffe0cb50ab4
==436==Using libbacktrace symbolizer.
==436==AddressSanitizer Init done
mount: /dev/sda1 mounted on /.

[1]+  Stopped                 mount -o remount,rw -v /
```

mrc0mmand:
Nevertheless, `detect_leaks=0` *fixes* the issue:

```
bash-5.0# export ASAN_OPTIONS=$ASAN_OPTIONS:detect_leaks=0
bash-5.0# mount -o remount,rw -v /
mount: /dev/sda1 mounted on /.
bash-5.0# echo $?
0
```

evverx:
Well, `asan_wrapper` is supposed to prepare all the services for ASan so I think it could reset `LD_PRELOAD` in `systemd-remount-fs.service` to get around that particular issue. Another option would be to try not to set `LD_PRELOAD` globally and add it to dbus only to fix https://github.com/systemd/systemd/issues/5004 but I'm not sure how stable this approach is.

mrc0mmand:
Thanks for the suggestion, it seems to be working!

```
...
TEST RUN: Basic systemd setup [OK]

TEST CLEANUP: Basic systemd setup
make: Leaving directory '/build/test/TEST-01-BASIC'
```

vagrant/vagrant-test-sanitizers.sh
```diff
@@ -18,7 +18,23 @@ export ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check
...
+export QEMU_TIMEOUT=600
+# Set QEMU_SMP to speed things up
+export QEMU_SMP=$(nproc)
+export TEST_NO_NSPAWN=1
+# Arch Linux requires booting with initrd, as all commonly used filesystems
+# are compiled in as modules
+export SKIP_INITRD=no
+
+rm -fr /var/tmp/systemd-test*
+exectask "TEST-01-BASIC_sanitizers" "make -C test/TEST-01-BASIC clean setup run clean-again"
```
evverx: I think it would be better to run the test in nspawn containers first as we do on Travis CI. Otherwise, we're going to be having a hard time figuring out what can possibly cause a kernel panic as in https://github.com/libbpf/libbpf/pull/29#issuecomment-482340365. In general it's risky to run PID1 under ASan+UBSan in VMs so the idea is to catch as many issues as possible in containers first. In that particular case, the test would have failed run via `nspawn` because on Arch the hybrid hierarchy is used in containers so we wouldn't have had to deal with that kernel panic at all.

evverx:
FWIW the good news is that I can't reproduce the bug I was talking about in https://github.com/systemd/systemd/pull/11592#issuecomment-458782477 on Arch, which means that that ugly workaround isn't necessary. The bad news is that [UBSan reports are in very bad shape](https://github.com/systemd/systemd/pull/12524).

vagrant/vagrant-test-sanitizers.sh
mrc0mmand: I actually thought it's the other way around - running potentially unstable PID 1 under nspawn, where it communicates with host's PID 1, sounded more dangerous to me. But with the kernel panic stuff it actually makes sense to run the nspawn test beforehand.

evverx:
It's somewhat tangential but looking at https://travis-ci.org/systemd/systemd/builds/530671639, I was trying to figure out why CentOS CI doesn't seem to have failed and it looks like in https://github.com/systemd/systemd-centos-ci/pull/99 `-werror` was accidentally removed.

mrc0mmand:
> It's somewhat tangential but looking at https://travis-ci.org/systemd/systemd/builds/530671639, I was trying to figure out why CentOS CI doesn't seem to have failed and it looks like in #99 `-werror` was accidentally removed.

It wasn't even there in the first place. For some reason I though `-Werror` it's on by default since systemd migrated to meson, but apparently I was wrong :-). It should definitely be there, though, thanks!

evverx:
I took a look at the build log to figure out whether the test is run in both "privileged" and userns containers and it seems that it's not possible to skim through the output of the test at https://ci.centos.org/job/systemd-ci-build/451/consoleFull. Is CentOS CI going to save it somewhere among "artifacts" once it's done?

mrc0mmand:
Each successfully finished (i.e. not aborted) job should finish with 

\************************************************************
 Landing page: https://ci.centos.org/job/systemd-ci-build/451/artifact//index.html
\************************************************************

message, which should provided at least a basic overview and access to the task logs. The lack of intermediate logs is kind of annoying, but that's unfortunately how the Jenkins artifacts work.

In this particular case the fail is caused by a flaky OOM policy test, which was filed once as https://github.com/systemd/systemd/issues/12425 but then closed as resolved...

evverx:
> but then closed as resolved...

That's the spirit :-)

The PR looks good to me. I guess I'll stress-test it anyway in https://github.com/systemd/systemd/pull/12524 once it's merged here.

evverx:
By the way, I'd merge the PR if I was 100% sure that it's OK. But I don't know what the workflow looks like here so I'll leave it to you :-)

I'm just spitballing :-) In theory if `systemd-networkd-test.py` wasn't so flaky it would be possible to  tweak `systemd-networkd.service` to turn off the settings incompatible with sanitizers and point it to the binary built with ASan+UBSan. It never disappoints:
```
=================================================================
==10565==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x613000007894 at pc 0x7fac7f037ee1 bp 0x7ffff0b32e10 sp 0x7ffff0b325b8
READ of size 32 at 0x613000007894 thread T0
    #0 0x7fac7f037ee0 in __interceptor_strndup /build/gcc/src/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:326
    #1 0x55dbbeb1e5ce in lease_parse_string ../src/libsystemd-network/sd-dhcp-lease.c:336
    #2 0x55dbbeb1e842 in lease_parse_domain ../src/libsystemd-network/sd-dhcp-lease.c:353
    #3 0x55dbbeb21786 in dhcp_lease_parse_options ../src/libsystemd-network/sd-dhcp-lease.c:591
    #4 0x55dbbeb16fb1 in parse_options ../src/libsystemd-network/dhcp-option.c:223
    #5 0x55dbbeb17291 in dhcp_option_parse ../src/libsystemd-network/dhcp-option.c:251
    #6 0x55dbbeaf2fb5 in client_handle_offer ../src/libsystemd-network/sd-dhcp-client.c:1297
    #7 0x55dbbeaf96b3 in client_handle_message ../src/libsystemd-network/sd-dhcp-client.c:1556
    #8 0x55dbbeafce8e in client_receive_message_raw ../src/libsystemd-network/sd-dhcp-client.c:1822
    #9 0x7fac7e4a39c7  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbe39c7)
    #10 0x7fac7e4aa142  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbea142)
    #11 0x7fac7e4ab0d7  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbeb0d7)
    #12 0x7fac7e4ab404  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbeb404)
    #13 0x55dbbe969908 in run ../src/network/networkd.c:113
    #14 0x55dbbe969b1f in main ../src/network/networkd.c:120
    #15 0x7fac7cbd9ce2 in __libc_start_main (/lib64/libc.so.6+0x23ce2)
    #16 0x55dbbe967bfd in _start (/home/vagrant/systemd/build/systemd-networkd+0x2b1bfd)
0x613000007894 is located 0 bytes to the right of 340-byte region [0x613000007740,0x613000007894)
allocated by thread T0 here:
    #0 0x7fac7f06f5a1 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:95
    #1 0x55dbbeafc634 in client_receive_message_raw ../src/libsystemd-network/sd-dhcp-client.c:1790
    #2 0x7fac7e4a39c7  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbe39c7)
    #3 0x7fac7e4aa142  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbea142)
    #4 0x7fac7e4ab0d7  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbeb0d7)
    #5 0x7fac7e4ab404  (/home/vagrant/systemd/build/src/shared/libsystemd-shared-242.so+0xbeb404)
    #6 0x55dbbe969908 in run ../src/network/networkd.c:113
    #7 0x55dbbe969b1f in main ../src/network/networkd.c:120
    #8 0x7fac7cbd9ce2 in __libc_start_main (/lib64/libc.so.6+0x23ce2)
SUMMARY: AddressSanitizer: heap-buffer-overflow /build/gcc/src/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:326 in __interceptor_strndup
```

mrc0mmand:
Honestly, I really wish the `systemd-networkd-tests.py` was more stable than it currently is. I wanted to make it part of the Arch Linux run in #83, but it's flaky even on baremetal, VM makes the matter worse. Also, even on baremetal it runs for like ~15 minutes, I imagine that in a VM under ASan+UBSan it would take at least twice as much time, so splitting the sanitizers run into a separate one would be definitely a necessity.

I'll merge this PR so you can stress it in the aforementioned PR, and I'll reopen the OOM policy issue, as the test is still flaky.

evverx:
It appears that CentOS CI runs the test in VMs even if it fails in containers (which is exactly what I was trying to avoid by tweaking the order of the tests). Other than that, everything seems to be fine.

evverx:
Regarding `systemd-networkd-tests.py`, I'd simply turn it off because nobody seems to be interested in make it (and `systemd-networkd` in general) work. Apparently new features barely covered by any testing whatsoever are more important and I'd personally have no problem with that :-) if I didn't have to waste a lot of time looking at the output of CentOS CI just to find out that it's `systemd-networkd.py` that fails more often than not.

evverx:
So that it wouldn't look like I'm complaining into the void :-) @yuwata @ssahani are you OK with turning `systemd-networkd-tests.py` off on CentOS CI?

yuwata:
Please do not turn off the test...

evverx:
@yuwata it's not that I want to turn it off :-). It's just that it's flaky. I'm wondering if anyone is going to address that flakiness?

yuwata:
You mean https://github.com/systemd/systemd/issues/12344 ?

evverx:
Well, I haven't kept track of what exactly fails there for a long time. I just look at the logs, notice that is's systemd-networkd again, sigh and move on. I believe @mrc0mmand have more advanced statistics.

yuwata:
OK. Anyway, I will try to tackle https://github.com/systemd/systemd/issues/12344 and #83, but maybe not possibly soon, sorry. Maybe next weekend or so.

evverx:
@yuwata could you turn off `test_bond_operstate` on CentOS in the meantime? I don't think there is any point in running it there given that we know it's flaky. It's hard to pay attention to CI when it's almost always red (which is probably why Ubuntu CI is ignored and I really hope it won't happen to CentOS CI)

evverx:
And there is nothing to be sorry about. We're all busy. I totally get it. That's basically why I have never reported that about 20 subtests are failing on my machine because unfortunately I don't have enough time to untangle why systemd-networkd fails to talk to the kernel and even if I did have the time, the bug report would most likely be open for eternity bothering me and everybody else :-)

ssahani:
@evverx thanks for pointing out . I will see why this is failing .

evverx:
@ssahani thank you!

I've just reread my comment about "new features barely covered by any testing". It sounds harsh and I apologize for the tone. What I was trying to say is that even though `systemd-networkd-tests.py` in fact covers a lot of features, some of them heavily rely on newer kernels and aren't tested on CentOS (which means that they are effectively not covered by testing on a regular basis). https://github.com/systemd/systemd-centos-ci/pull/83 will help to address that.

mrc0mmand:
> It appears that CentOS CI runs the test in VMs even if it fails in containers (which is exactly what I was trying to avoid by tweaking the order of the tests). Other than that, everything seems to be fine.

I accidentally dropped that part during debugging, should be fixed in #105.

evverx:
@mrc0mmand thank you!

FYI I've just hooked up the systemd repository to Azure Pipelines to move a couple of stages from Travis CI (in case you're interested I'm experimenting with it in https://github.com/systemd/systemd/pull/12542). What's more important though it that we basically have 10 Ubuntu Xenial VMs waiting to be unleashed, which might be useful if we decide to run `systemd-networkd` under ASan+UBsan. It's possible to play with them by opening a PR and tweaking https://github.com/systemd/systemd/blob/master/azure-pipelines.yml.

evverx:
Speaking of Ubuntu Xenial, it ships with Python 3.5 (where f-strings aren't supported) so [`systemd-networkd-tests.py` fails as soon as it starts there with](https://dev.azure.com/evvers/systemd-systemd/_build/results?buildId=27&view=logs&jobId=56fae37d-e638-5fac-6ab9-ca4fcdfee404&taskId=e3a12e69-71ef-587d-d243-f543ea61fe68&lineStart=3109&lineEnd=3115&colStart=1&colEnd=28)
```
Also=systemd-networkd-wait-online.service
+ cd test/test-network
+ ./systemd-networkd-tests.py
  File "./systemd-networkd-tests.py", line 143
    dnsmasq_command = f'dnsmasq -8 /var/run/networkd-ci/test-dnsmasq-log-file --log-queries=extra --log-dhcp --pid-file=/var/run/networkd-ci/test-test-dnsmasq.pid --conf-file=/dev/null --interface=veth-peer --enable-ra --dhcp-range=2600::10,2600::20,{lease_time} --dhcp-range=192.168.5.10,192.168.5.200,{lease_time} -R --dhcp-leasefile=/var/run/networkd-ci/lease --dhcp-option=26,1492 --dhcp-option=option:router,192.168.5.1 --dhcp-option=33,192.168.5.4,192.168.5.5 --port=0 ' + additional_options
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           ^
SyntaxError: invalid syntax
```
It's not that it's impossible to install a newer version of python. It's just maybe it would be better to stick to less advanced features available on all platforms `systemd` targets (if the ultimate goal is to make it easier for people to run the test of course). And how do the f-strings ever work on CentOS? :-)

mrc0mmand:
> And how do the f-strings ever work on CentOS? :-)

That's because I needed Python 3.5 due to meson requirements and CentOS ships (or more specifically - used to ship) with Python 3.4 in base repos and Python 3.6 in EPEL. Thus, the easiest path for me was to just use the Python 3.6 package, where the f-strings work as expected :-)

https://github.com/systemd/systemd-centos-ci/blob/master/agent/bootstrap.sh#L48-L50

ssahani:
```
Also=systemd-networkd-wait-online.service
+ cd test/test-network
+ ./systemd-networkd-tests.py
  File "./systemd-networkd-tests.py", line 143
    dnsmasq_command = f'dnsmasq -8 /var/run/networkd-ci/test-dnsmasq-log-file --log-queries=extra --log-dhcp --pid-file=/var/run/networkd-ci/test-test-dnsmasq.pid --conf-file=/dev/null --interface=veth-peer --enable-ra --dhcp-range=2600::10,2600::20,{lease_time} --dhcp-range=192.168.5.10,192.168.5.200,{lease_time} -R --dhcp-leasefile=/var/run/networkd-ci/lease --dhcp-option=26,1492 --dhcp-option=option:router,192.168.5.1 --dhcp-option=33,192.168.5.4,192.168.5.5 --port=0 ' + additional_options
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           ^
SyntaxError: invalid syntax
```

Should we fix this @evverx

mrc0mmand:
I'd fix that, as it's going to be a while before Python 3.6 is present in all environments we use for CI's, so this would allow us run the systemd-networkd testsuite without additional kludges. Thankfully, falling back to `'{}'.format(x)` is not *that* bad, so the change shouldn't affect the readability of the code.

PR #108, opened 2019-05-12
vagrant: don't wait indefinitely when VM reload goes south

mrc0mmand:
Accidentally found by systemd/systemd#12542


mrc0mmand:
The OOM policy fail is tracked in https://github.com/systemd/systemd/issues/12425

PR #110, opened 2019-05-17
tree-wide: generate man pages by default

mrc0mmand:
When systemd/systemd#12519 gets merged, man pages will be built on-demand
only. Let's enable the man page generation explicitly, so we can test it
even after this change.


keszybz:
LGTM.

If the goal is to build everything, `-Dhtml=true` could be added too.

mrc0mmand:
> If the goal is to build everything, `-Dhtml=true` could be added too.

Good point, added, thanks!

PR #114, opened 2019-05-23
vagrant: add sanitizers/clang Vagrantfile

mrc0mmand:


evverx:
Currently `asan_wrapper` is incompatible with systemd built with clang so I'd skip TEST-01-BASIC for now. It'd be great to make it work both with gcc and clang though (maybe later :-)).

mrc0mmand:
> Currently `asan_wrapper` is incompatible with systemd built with clang so I'd skip TEST-01-BASIC for now. It'd be great to make it work both with gcc and clang though (maybe later :-)).

Ah, thank you, I was under impression I messed something up :-) I'll try to play around with it for a while to see what can be done about that.

Anyway, I had to bump the image size from 1 GB to 2 GB, because with clang the setup phase was running out of memory:

```
[TASK START] Thu 23 May 2019 06:28:02 AM PDT
make: Entering directory '/build/test/TEST-01-BASIC'
TEST CLEANUP: Basic systemd setup
TEST SETUP: Basic systemd setup
0+0 records in
0+0 records out
0 bytes copied, 0.000341529 s, 0.0 kB/s
Checking that no-one is using this disk right now ... OK

Disk /dev/loop0: 1000 MiB, 1048576000 bytes, 2048000 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Created a new DOS disklabel with disk identifier 0xa582d08e.
/dev/loop0p1: Created a new partition 1 of type 'Linux' and of size 950 MiB.
/dev/loop0p2: Created a new partition 2 of type 'Linux' and of size 49 MiB.
/dev/loop0p3: Done.

New situation:
Disklabel type: dos
Disk identifier: 0xa582d08e

Device       Boot   Start     End Sectors  Size Id Type
/dev/loop0p1         2048 1947647 1945600  950M 83 Linux
/dev/loop0p2      1947648 2047999  100352   49M 83 Linux

The partition table has been altered.
Calling ioctl() to re-read partition table.
Syncing disks.
+ DESTDIR=/var/tmp/systemd-test.fncGFy/root
+ /sbin/ninja -C /build/build install
ninja: Entering directory `/build/build'
[1/392] Generating version.h with a custom command.
[1/2] Installing files.
  File "/usr/lib/python3.7/shutil.py", line 122, in copyfile
    copyfileobj(fsrc, fdst)
  File "/usr/lib/python3.7/shutil.py", line 82, in copyfileobj
    fdst.write(buf)
OSError: [Errno 28] No space left on device

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.7/site-packages/mesonbuild/mesonmain.py", line 122, in run
    return options.run_func(options)
  File "/usr/lib/python3.7/site-packages/mesonbuild/minstall.py", line 514, in run
    installer.do_install(datafilename)
  File "/usr/lib/python3.7/site-packages/mesonbuild/minstall.py", line 343, in do_install
    self.install_targets(d)
  File "/usr/lib/python3.7/site-packages/mesonbuild/minstall.py", line 450, in install_targets
    self.do_copyfile(fname, outname)
  File "/usr/lib/python3.7/site-packages/mesonbuild/minstall.py", line 257, in do_copyfile
    shutil.copyfile(from_file, to_file)
  File "/usr/lib/python3.7/shutil.py", line 122, in copyfile
    copyfileobj(fsrc, fdst)
OSError: [Errno 28] No space left on device
```

mrc0mmand:
Okay, the out-of-space error is most likely due to the additonal man and HTML pages we recently added to the compilation process. I'm somewhat inclined to drop both from the sanitizers run, as building them during the *standard* runs should be enough.

mrc0mmand:
Also, to make the ASan wrapper work with clang, we'd need to compile the compile-rt[0] part of LLVM with `-DCOMPILER_RT_BUILD_SHARED_ASAN=ON`, which should then generate a `libasan.so*` DSO usable within the wrapper. However, doing this in a reliable way might be a PITA, as we'd need to checkout the correct revision (probably using LLVM SVN tags[1]) and only the repo checkout takes *ages*, let alone the compilation itself...

I'm not sure if it's the ideal way, but I think that's how it's done in oss-fuzz.

[0] https://llvm.org/svn/llvm-project/compiler-rt/
[1] https://llvm.org/svn/llvm-project/compiler-rt/tags

mrc0mmand:
Apparently, on Arch Linux, there's a `compiler-rt` package which should have all these libraries already compiled:

```
# pacman -Ql compiler-rt | grep asan
compiler-rt /usr/include/sanitizer/asan_interface.h
compiler-rt /usr/include/sanitizer/hwasan_interface.h
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-i386.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-i386.so
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-preinit-i386.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-preinit-x86_64.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-x86_64.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-x86_64.a.syms
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan-x86_64.so
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan_cxx-i386.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan_cxx-x86_64.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.asan_cxx-x86_64.a.syms
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.hwasan-x86_64.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.hwasan-x86_64.a.syms
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.hwasan-x86_64.so
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.hwasan_cxx-x86_64.a
compiler-rt /usr/lib/clang/8.0.0/lib/linux/libclang_rt.hwasan_cxx-x86_64.a.syms
compiler-rt /usr/lib/clang/8.0.0/share/asan_blacklist.txt
compiler-rt /usr/lib/clang/8.0.0/share/hwasan_blacklist.txt
```

This might make this much easier.

mrc0mmand:
@evverx so, after a short trial & error session, the `TEST-01-BASIC` test works even under clang with sanitizers.

What's need to be changed in systemd/test/test-functions:
1. make sure the `compiler-rt` package is installed (this applies for Arch Linux, other distributions may not even have such package and would have to compile the library from the LLVM repos, see above)
2. bump `local _size` from 500 to 1000 MB; even with man pages disabled 1 GB (2 * `_size` when [running under ASan without stripping binaries](https://github.com/systemd/systemd/blob/master/test/test-functions#L483-L485)) is not enough
3. install the runtime LLVM ASan DSO to the test image (`inst_library $(find /usr/lib/clang -name libclang_rt.asan-x86_64.so | sed 1q)`)
4. change the find pattern for `PATH_TO_ASAN` (I used `sed -i 's/\*libasan\*/libclang_rt.asan-x86_64.so/' test/test-functions`, but it needs a nicer solution)

For all this to work as expected we need to somehow detect we were compiled with clang, to preload the correct ASan DSO.

evverx:
@mrc0mmand you're unstoppable :-) I thought we'd just start from running the unit tests built with clang under ASan+UBSan so that we could go ahead and add `-e container=docker` in https://github.com/systemd/systemd/pull/12617 :-) Given that familiarizing `asan_wrapper` with clang will most likely take a while (and need some memory leaks and UB introduced temporarily to make sure it works) I'd move this part in a new PR against the systemd repository where you could experiment with it (on Travis CI for example by resurrecting the Xenial phase removed in https://github.com/systemd/systemd/commit/04ea08b5a486159c4f8e1ce0aa645b35fa66c822#diff-354f30a63fb0907d4ad57269548329e3 where clang would be used instead of gcc).

> For all this to work as expected we need to somehow detect we were compiled with clang, to preload the correct ASan DSO.

I think it should be possible to figure out which libraries `systemd` linked against with `ldd`, that is something like:
```sh
asan_lib=$(ldd ... | grep asan)
if ! asan_lib
      echo "No can do"
      exit 1
elif /* gcc-lib here */

elif /* clang-lib here */

else
     echo "Have no idea what $asan_lib is"
     exit 1
fi
```
should probably suffice.

evverx:
Plus as far as I can remember to get meaningful reports llvm-symbolizer should be installed too and *_SYMBOLIZER_PATH should be set accordingly if necessary. Without it backtraces are kind of hard to decipher.

evverx:
> I'd move this part to a new PR against the systemd repository where you could experiment with it

Another option would be to keep experimenting here by pointing CentOS CI to your fork of the systemd repository where you could change `test-functions` and add UB and memory leaks. I'm not sure what's more convenient :-)

mrc0mmand:
I was just curious how hard it would actually be to make the ASan wrapper work with both gcc and clang :-). I'll drop the clang-specific shenanigans and disable the `TEST-01-BASIC` so we can merge this.

evverx:
> I'll drop the clang-specific shenanigans and disable the TEST-01-BASIC so we can merge this.

@mrc0mmand could you skip TEST-01-BASIC only when systemd is built with clang? It seems to be risky to turn it off completely on CentOS CI given that it's off everywhere else.

mrc0mmand:
> > I'll drop the clang-specific shenanigans and disable the TEST-01-BASIC so we can merge this.
> 
> @mrc0mmand could you skip TEST-01-BASIC only when systemd is built with clang? It seems to be risky to turn it off completely on CentOS CI given that it's off everywhere else.

Ugh, thanks, made this change before my first coffee, so I completely missed the fact the `vagrant-test-sanitizers.sh` script is shared between both gcc and clang.

mrc0mmand:
Looks like everything runs as and when expected https://ci.centos.org/job/systemd-ci-build/526/artifact//index.html

Will merge this now, edit the appropriate Jenkins job, so it's executed in the same run as the gcc sanitizer run. Thanks for the help!

PR #115, opened 2019-05-24
vagrant: bump the nspawn timeout to 15 minutes

mrc0mmand:
The amount of unit tests in TEST-24-UNIT-TESTS rises slowly, but
steadily, and it started triggering the 10m timeout in systemd-nspawn.
Let's bump it to 15m

cc @yuwata


yuwata:
Thanks!

PR #123, opened 2019-05-29
vagrant: install wireguard-arch package

### LINKED ISSUE #122: wireguard doesn't seem to be installed on Arch when systemd-networkd-test.py is run
To judge from "modprobe: FATAL: Module wireguard not found in directory /lib/modules/5.1.4-arch1-1-ARCH" I've just found in https://ci.centos.org/job/systemd-pr-build-vagrant/2065/artifact//systemd-centos-ci/artifacts_w3PKi7/vagrant-logs.FM4/vagrant-arch-testsuite.SYc/systemd-networkd-tests.py_PASS.log the wireguard kernel module wasn't installed and all the tests covering wireguard were skipped.

Originally I discovered it in https://github.com/systemd/systemd-centos-ci/pull/117 but I didn't want to bury it there because I kind of expect that discussion to be lengthy :-)

mrc0mmand:
This should allow us to run wireguard tests from the systemd-networkd
testsuite

Fixes #122

### DIFF
--- a/vagrant/boxes/Vagrantfile_archlinux_systemd
+++ b/vagrant/boxes/Vagrantfile_archlinux_systemd
@@ -59,6 +59,6 @@ Vagrant.configure("2") do |config|
     # Note: openbsd-netcat in favor of gnu-netcat is used intentionally, as
     #       the GNU one doesn't support -U option required by test/TEST-12-ISSUE-3171
     pacman --needed --noconfirm -S net-tools strace openbsd-netcat busybox e2fsprogs quota-tools \
-        dnsmasq automake make dhclient rsync qemu socat
+        dnsmasq automake make dhclient rsync qemu socat wireguard-arch
   SHELL
 end


evverx:
Today I learned that images are stored in a cache of some kind :-) I'm wondering how long it usually takes for an image to end up there and whether it happens automatically.

By the way, I tried to download Arch from the cache but it took so long (because I suspect it isn't behind a CDN) so I didn't even wait for it to finish.

mrc0mmand:
I introduced a cache, because the upgrade and package installation took ~10 minutes in each run, which could have been spent by something more useful :-). The cache mechanism is not particularly nice, as it's, unfortunately, full of kludges (see https://github.com/systemd/systemd-centos-ci/blob/master/vagrant/vagrant-make-cache.sh), but it really speeds things up.

The cache itself is periodically updated by https://ci.centos.org/job/systemd-centos-ci-vagrant-make-cache/ every three days, but when I do changes like this, I run the sync script as part of the CI run, so the cache is updated immediately. Again, this is pretty ugly, as it may affect "production" runs, but thankfully these changes are rare and usually simple as this one.

> By the way, I tried to download Arch from the cache but it took so long (because I suspect it isn't behind a CDN) so I didn't even wait for it to finish.

Yeah, downloading the image directly is woefully slow, as it's not behind any CDN, but it was the most convenient way for CI purposes. It makes the local runs a bit more complicated - in such cases I usually copy&paste the provision script from the `vagrant/boxes/Vagrantfile_archlinux_systemd` to the respective Vagrantfile from `vagrant/Vagrantfiles`, so the upgrade and package installation is done locally as well. It's slow, but still faster than waiting for the cached image to download.

mrc0mmand:
Also, looks like the wireguard tests are finally being run:

```
test_wireguard (__main__.NetworkdNetDevTests) ... RTNETLINK answers: Invalid argument
RTNETLINK answers: Invalid argument
RTNETLINK answers: Invalid argument
RTNETLINK answers: Invalid argument

ok
```

PR #124, opened 2019-05-31
debug-only: udev debug

mrc0mmand:
Let's attempt to run systemd/systemd#12700 with `udev.log_priority=debug` (CentOS CI) and systemd/systemd#12708 with a simple find to locate `99-default.link` (CentOS CI - Arch in KVM).

I really need to make this configurable per-PR.


yuwata:
Thank you!

PR #128, opened 2019-05-31
Introduce pre/post tasks

mrc0mmand:
After recent @yuwata's struggles with getting proper debug logs for certain PRs without too much hassle, I tried to introduce a concept of pre and post tasks. Basically, they should allow you to tweak the build environment before and after each phase to make the debugging much easier without creating any additional PRs in this repository and trying to point the Jenkins job to the right place.

To better explain how this should all work, I wrote a short documentation draft. If this all makes sense, I'll do something similar to the Vagrant jobs as well.

cc @evverx

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -72,6 +72,10 @@ pushd systemd
 
 git_checkout_pr "${1:-""}"
 
+# If a pre-bootstrap script exists in the root of the systemd repo, source it
+test -e .centosci-pre-bootstrap && source .centosci-pre-bootstrap
+
+MESON_ARGS="${MESON_ARGS:-""}"
 # It's impossible to keep the local SELinux policy database up-to-date with
 # arbitrary pull request branches we're testing against.
 # Disable SELinux on the test hosts and avoid false positives.
@@ -97,7 +101,7 @@ systemctl disable firewalld
                 -Dnobody-user=nfsnobody \
                 -Dnobody-group=nfsnobody \
                 -Dman=true \
-                -Dhtml=true
+                -Dhtml=true $MESON_ARGS
     ninja-build -C build
 ) 2>&1 | tee "$LOGDIR/build.log"
 
@@ -148,6 +152,9 @@ fi
 grubby --args="user_namespace.enable=1" --update-kernel="$(grubby --default-kernel)"
 grep "user_namespace.enable=1" /boot/grub2/grub.cfg
 echo "user.max_user_namespaces=10000" >> /etc/sysctl.conf
+
+# If a post-bootstrap script exists in the root of the systemd repo, source it
+test -e .centosci-post-bootstrap && source .centosci-post-bootstrap
 
 echo "-----------------------------"
 echo "- REBOOT THE MACHINE BEFORE -"
--- a/agent/testsuite.sh
+++ b/agent/testsuite.sh
@@ -32,6 +32,10 @@ set +e
 ### TEST PHASE ###
 cd systemd
 
+
+# If a pre-testsuite script exists in the root of the systemd repo, source it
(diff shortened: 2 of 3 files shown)


evverx:
@mrc0mmand I think in general it makes sense but I have to say I'm not a fan of scattering configuration affecting CI across different places mainly because it's too easy to forget to remove kludges and when that happens you need to jump between repositories to try to figure out why it works the way it works. If we go down that road we'll have to somehow prevent PRs where the tasks are used from merging.

Another option would be to merge CentOS CI into the systemd repository.

mrc0mmand:
Yeah, that's a good point. Truth be told, I came up with this only to present a possible alternative for the current *PR debug mess*, in case you and other systemd developer find it too obnoxious. If you're okay with how things work now (i.e. pinging me when something goes haywire), we can definitely throw this PR away :-)

evverx:
Well, I'm not exactly sure what that mess you're referring to is :-), what the issue with debug logs you pointed out in the description was and why it wasn't possible to set the log level to `debug` explicitly (in `systemd-udevd` or `systemd-networkd` I assume) and then look at the journal (which can be found among the "artifacts" as far as I know). I wouldn't throw this PR away before figuring out what we're trying to accomplish :-)

mrc0mmand:
The most recent case I think of is https://github.com/systemd/systemd/pull/12700#issuecomment-497596319 where setting the `udev.log_priority=debug` can't be done without tweaking the CentOS CI side in a separate PR and linking these two together. Similarly, you may need to enable debug logging for certain part of system, but only temporary. In such situations the task script may come in handy. The question is if it's worth it, given how often something like this is needed :-)

evverx:
> The most recent case I think of is systemd/systemd#12700 (comment) where setting the udev.log_priority=debug can't be done without tweaking the CentOS CI side in a separate PR and linking these two together.

I took a look at the code and it seems that it would have been possible to just call something like `log_set_max_level(LOG_DEBUG);` in `run` unconditionally (as far as I can tell, that's exactly what `udev.log_priority=debug` usually leads to)

> Similarly, you may need to enable debug logging for certain part of system, but only temporary. 

My understanding is that anything related to systemd (the code, the unit files, the default config files and so on) can be temporarily tweaked to get almost anything anyone ever needs without changing anything on the CentOS CI side.

> The question is if it's worth it

I think it depends. On a somewhat unrelated note, for example, in https://github.com/systemd/systemd/pull/8207#issuecomment-492501961 `libzstd` should be installed to get any meaningful results (and if meson didn't default to it in meson.build something like `-Dzstd=true` would have to be passed as well). Looks like the "tasks" should somehow help to cover use cases like this.

evverx:
> Looks like the "tasks" should somehow help to cover use cases like this.

Having said that, if `meson` is run as `root` there, it should be possible to temporarily install additional packages via `run_command` :-) so I'm inclined to say that I'd leave everything as is.

Anyway, I'd wait for someone else to weigh in here because it could be that for some reason the debugging techniques I mentioned (let's call them that :-) are somewhat unacceptable.

evverx:
@mrc0mmand it's absolutely unrelated but I came across https://lists.freedesktop.org/archives/systemd-devel/2019-May/042791.html the other day and I think it would be great if there was a way for new contributors to figure out what to do if they have any questions. Maybe something like "Cent OS CI (Arch in KVM) (reach out to @mrc0mmand if you have any questions)" in the status message would suffice. What do you think?

mrc0mmand:
That would definitely make sense. I tried to do something like that [here](https://ci.centos.org/view/systemd/) but given that this page isn't linked anywhere, it's kind of pointless. You can always click on the avatar next to the status message, which should lead to my profile, but I admit it's not too obvious.

PR #132, opened 2019-06-18
vagrant: run slow tests under sanitizers

mrc0mmand:
Another attempt in running the slow tests (`-Dslow-tests=true`) under ASan & UBSan. Let's see what else is broken (if anything) since https://github.com/systemd/systemd/pull/12824 was merged.

### DIFF
--- a/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-clang
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-clang
@@ -93,6 +93,7 @@ Vagrant.configure("2") do |config|
           --werror \
           -Dc_args='-g -O0 -ftrapv' \
           -Dtests=unsafe \
+          -Dslow-tests=true \
           -Dinstall-tests=true \
           -Ddbuspolicydir=/usr/share/dbus-1/system.d \
           -Dman=false \
--- a/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-gcc
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-gcc
@@ -87,6 +87,7 @@ Vagrant.configure("2") do |config|
           --werror \
           -Dc_args='-g -O0 -ftrapv' \
           -Dtests=unsafe \
+          -Dslow-tests=true \
           -Dinstall-tests=true \
           -Ddbuspolicydir=/usr/share/dbus-1/system.d \
           -Dman=false \
--- a/vagrant/vagrant-test-sanitizers.sh
+++ b/vagrant/vagrant-test-sanitizers.sh
@@ -25,7 +25,7 @@ if _clang_asan_rt_name="$(ldd build/systemd | awk '/libclang_rt.asan/ {print $1;
 fi
 
 # Run the internal unit tests (make check)
-exectask "ninja-test_sanitizers" "meson test -C build --print-errorlogs --timeout-multiplier=3"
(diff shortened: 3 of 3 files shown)


mrc0mmand:
@evverx is it worth bumping the `--timeout-multiplier` even more (i.e. 10+) for these ultra-slow tests?

```
369/773 test-hashmap                            TIMEOUT 900.07 s 

--- command ---
SYSTEMD_LANGUAGE_FALLBACK_MAP='/build/src/locale/language-fallback-map' SYSTEMD_KBD_MODEL_MAP='/build/src/locale/kbd-model-map' PATH='/build/build:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl' /build/build/test-hashmap
--- stderr ---
/* test_hashmap_copy */
/* test_hashmap_get_strv */
/* test_hashmap_move_one */
/* test_hashmap_move */
/* test_hashmap_replace */
/* test_hashmap_update */
/* test_hashmap_put */
/* test_hashmap_remove */
/* test_hashmap_remove2 */
/* test_hashmap_remove_value */
/* test_hashmap_remove_and_put */
/* test_hashmap_remove_and_replace */
/* test_hashmap_ensure_allocated */
/* test_hashmap_foreach */
/* test_hashmap_foreach_key */
/* test_hashmap_contains */
/* test_hashmap_merge */
/* test_hashmap_isempty */
/* test_hashmap_get */
/* test_hashmap_get2 */
/* test_hashmap_size */
/* test_hashmap_many (slow) */
trivial_hashmap_ops 1048576 <= 1973790 * 0.8 = 1.57903e+06
test took 10min 29.212695s
crippled_hashmap_ops 16384 <= 30840 * 0.8 = 24672
test took 47.226263s
/* test_hashmap_free (slow, 1048576 entries) */
string_hash_ops test took 2min 12.703746s
-------

742/773 fuzz-unit-file:oss-fuzz-11569:address   TIMEOUT 300.04 s 

--- command ---
/usr/bin/env /build/build/fuzz-unit-file:address /build/test/fuzz/fuzz-unit-file/oss-fuzz-11569
--- stdout ---
/build/test/fuzz/fuzz-unit-file/oss-fuzz-11569... 
--- stderr ---
Found container virtualization none.
-------
```

evverx:
`strict_string_checks` makes some linear operations somewhat exponential and, apparently, it's not something `test-hashmap` is particularly happy about. I think for now it would be better to somehow exclude the test before running the test suite. If `fuzz-11569` is a "timeout" issue that wasn't triggered by an infinite loop I think it can safely be excluded as well. Generally I don't think it makes sense to keep the files triggering timeouts because the environment on OSS-Fuzz is different and if it's not recreated on Travis CI and CentOS CI by passing the same flags to compiler, the same ASAN_OPTIONS and so on, it doesn't make much sense to try to catch those regressions.

evverx:
Another option would be not to turn on the slow tests under ASan+UBsan :-) As far as I can remember some of those tests like `test-barrier` have hard-coded timeouts the global timeout doesn't affect so they will most likely be flaky.

mrc0mmand:
> Another option would be not to turn on the slow tests under ASan+UBsan :-) As far as I can remember some of those tests like `test-barrier` have hard-coded timeouts the global timeout doesn't affect so they will most likely be flaky.

Ah, that something I didn't know. In that case, I think it would be better to leave the sanitizer job as it is, thanks :-)

PR #141, opened 2019-07-06
agent/vagrant: extend timeout for other integration tests

### LINKED ISSUE #140: systemd-networkd-tests.py timeouts
After https://github.com/systemd/systemd/pull/12968, systemd-networkd-tests.py randomly timeouts. See https://github.com/systemd/systemd/pull/12972 and https://github.com/systemd/systemd/pull/12973.

@mrc0mmand Could you extend the timeout?

mrc0mmand:
This change is relevant mostly for the systemd-network testsuite, which
keeps growing and the original 30m timeout was simply not enough.

Fixes #140

### DIFF
--- a/agent/testsuite.sh
+++ b/agent/testsuite.sh
@@ -89,7 +89,7 @@ TEST_LIST=(
 )
 
 for t in "${TEST_LIST[@]}"; do
-    exectask "${t##*/}" "timeout 30m ./$t"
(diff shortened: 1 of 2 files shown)


mrc0mmand:
Let's skip the CI in this case, as the change *shouldn't* break anything, and it should prevent additional timeouts from being triggered in the wild.

Issue #137, opened 2019-06-21
Revert #136 once the gcc9 performance issues are resolved


mrc0mmand:
See:
* #136
* https://github.com/systemd/systemd/pull/12856
* https://bugzilla.redhat.com/show_bug.cgi?id=1722949

mrc0mmand:
@evverx I filed the GCC issue to upstream bug tracker with some statistics (possibly explaining the performance issues), if you're interested: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91101

evverx:
@mrc0mmand thank you. In the meantime, I think it should be possible to keep rolling forward by reverting that commit where the repositories were pinned and either turning off the stage where the code is built with gcc+ASan or by stopping passing `detect_stack_use_after_return=1` there. It should be safe because it's already covered by the "clang" stage.

mrc0mmand:
@evverx that would indeed make sense; and given the response from the gcc guys, it looks like there might be some additional tweaks necessary https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91101#c15

mrc0mmand:
Fixed in #146

PR #142, opened 2019-07-08
Temporarily disable `detect_stack_use_after_return` check for gcc runs

mrc0mmand:
Let's temporarily disable `disable detect_stack_use_after_return` check for gcc runs until https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91101 is resolved.

See #137

### DIFF
--- a/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-clang
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-clang
@@ -14,8 +14,7 @@ Vagrant.configure("2") do |config|
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://vagrantcloud.com/search.
   # Use our updated & cached Vagrant box (see vagrant/vagrant-make-cache.sh)
-  config.vm.box = "archlinux_systemd"
-  config.vm.box_url = "http://artifacts.ci.centos.org/systemd/vagrant_boxes/archlinux_systemd"
+  config.vm.box = "generic/arch"
 
   # Disable automatic box update checking. If you disable this, then
   # boxes will only be checked for updates when the user runs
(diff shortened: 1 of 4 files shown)


mrc0mmand:
Note: the patch from systemd/systemd#12856 is necessary to build systemd with gcc9

mrc0mmand:
I'll leave it as it is, for now, due to several reasons:

1) According to the resolution of the [gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91101 i) the issue is in the systemd's codebase, so it needs to be investigated first
2) I'd like to wait for gcc 9.1.1, because as @yuwata found out in #12856, some issues are gcc 9.1.0 specific
3) The systemd-networkd testsuite is unfortunately failing and I'm not sure if the fails are not false positives. However, the testsuite fails on both gcc AND clang, probably caused by newer kernel. (cc @yuwata)

gcc: https://ci.centos.org/job/systemd-ci-build/713/artifact//artifacts_all/artifacts_nahcK8/vagrant-logs.t08/vagrant-arch-sanitizers-gcc-testsuite.3oR/systemd-networkd_sanitizers_FAIL.log
clang: https://ci.centos.org/job/systemd-ci-build/713/artifact//artifacts_all/artifacts_Amaz7A/vagrant-logs.FGn/vagrant-arch-sanitizers-clang-testsuite.IK5/systemd-networkd_sanitizers_FAIL.log

evverx:
@mrc0mmand since it was me who suggested keeping rolling forward, just to clarify, I wasn't sure how long Arch folks keep the repositories we pinned and when I saw that the issue was opened 2 weeks ago I started to panic. It they're not going to vanish into thin air any time soon I think it's OK to wait. Though, given the pace, I'd wait for a month at most so that new issues wouldn't pile up too long.

evverx:
Regarding the issues pointed out in the gcc bug, could you open an issue to kick off the descussion on the systemd bug tracker?

mrc0mmand:
> @mrc0mmand since it was me who suggested keeping rolling forward, just to clarify, I wasn't sure how long Arch folks keep the repositories we pinned and when I saw that the issue was opened 2 weeks ago I started to panic. It they're not going to vanish into thin air any time soon I think it's OK to wait. Though, given the pace, I'd wait for a month at most so that new issues wouldn't pile up too long.

Even though the Arch archive is pretty stable (I think the "live" snapshots go all the way back to 2012, the rest is on the Archive.org), in case of Arch I like the idea we test with the latest packages, because, as you can see, the newer gcc and kernel cause a fair amount of potential issues which could have been already caught and resolved. On the other hand, I don't want to end up with a CI, which tests all Arch Linux packages :-)

mrc0mmand:
> Regarding the issues pointed out in the gcc bug, could you open an issue to kick off the descussion on the systemd bug tracker?

Will do.

mrc0mmand:
Reference issue for the gcc/systemd bug: systemd/systemd#12997

mrc0mmand:
This PR was superseded by #144. I'll prepare a new one to test how the current codebase will fare with gcc 9.1.x (especially systemd-networkd).

PR #144, opened 2019-07-10
tree-wide: use -O1 -fno-omit-frame-pointer when compiling systemd

mrc0mmand:
After @jakubjelinek's hint (thanks a lot for it) in https://github.com/systemd/systemd/issues/12997 I went through the sanitizer documentations again and learned something I managed to miss multiple times in the past. Originally, I put `-O0` everywhere as I thought other optimization levels would interfere with the sanitizers, causing potential false positives, break stack unwinding, etc. To my surprise, all sanitizer documentations recommend using at least `-O1` but with `-fno-omit-frame-pointer` to keep the stack unwinding untouched.

That being the case, let's use `-O1` for (almost) all systemd builds, which will have some performance impact during compilation, but this should be evened up (most likely several times) with a better runtime of the tests.

I noticed that `-Og` is basically `-O1` with some other debugging-interfering options turned off, but as I have no experience with it in combination with sanitizers, I decided to stick with `-O1` from the documentation.

@evverx any thoughts?

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -90,7 +90,7 @@ systemctl disable firewalld
 (
     # Make sure we copy over the meson logs even if the compilation fails
     trap "[[ -d $PWD/build/meson-logs ]] && cp -r $PWD/build/meson-logs '$LOGDIR'" EXIT
-    meson build -Dc_args='-g -O0 -ftrapv' \
+    meson build -Dc_args='-g -O1 -fno-omit-frame-pointer -ftrapv' \
                 --werror \
                 -Dslow-tests=true \
                 -Dtests=unsafe \
--- a/vagrant/Vagrantfiles/Vagrantfile_arch
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch
@@ -91,7 +91,7 @@ Vagrant.configure("2") do |config|
     # Build phase
     meson build \
           --werror \
-          -Dc_args='-g -O0 -ftrapv' \
+          -Dc_args='-g -O1 -fno-omit-frame-pointer -ftrapv' \
           -Dslow-tests=true \
           -Dtests=unsafe \
           -Dinstall-tests=true \
--- a/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-clang
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-clang
@@ -99,7 +99,7 @@ Vagrant.configure("2") do |config|
 
     meson build \
           --werror \
-          -Dc_args='-g -O0 -ftrapv' \
+          -Dc_args='-g -O1 -fno-omit-frame-pointer -ftrapv' \
           -Dtests=unsafe \
           -Dinstall-tests=true \
           -Ddbuspolicydir=/usr/share/dbus-1/system.d \
--- a/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-gcc
+++ b/vagrant/Vagrantfiles/Vagrantfile_arch-sanitizers-gcc
@@ -93,7 +93,7 @@ Vagrant.configure("2") do |config|
     # Sanitizer (UBSan)
     meson build \
           --werror \
-          -Dc_args='-g -O0 -ftrapv' \
+          -Dc_args='-g -O1 -fno-omit-frame-pointer -ftrapv' \
           -Dtests=unsafe \
           -Dinstall-tests=true \
           -Ddbuspolicydir=/usr/share/dbus-1/system.d \


evverx:
@mrc0mmand I'm not sure where `-O0` originally came from. It was introduced back in 2016 in e50a072cdf3214fe115b74 and my guess is that it was supposed to prevent the compiler from compiling out anything (which I think in a way makes sense). On the other hand, with sanitizers it's certainly unbearable and, for example, when the fuzzers are built `-01` is passed explicitly: https://github.com/systemd/systemd/blob/6f209cb01d3d85d9073b9d6ca6f5dc3ae49c37dd/tools/oss-fuzz.sh#L13. I'm not sure what's going on Travis CI and Azure Pipelines though.

evverx:
@mrc0mmand could you try to turn on the slow tests under ASan+UBsan while you're at it to see how they fare with `-O1`?

mrc0mmand:
>  I'm not sure what's going on Travis CI and Azure Pipelines though.

Looks like both [xenial.sh](https://github.com/systemd/systemd/blob/master/travis-ci/managers/xenial.sh) and [debian.sh](https://github.com/systemd/systemd/blob/master/travis-ci/managers/debian.sh) use the default flags (as I don't see any `CFLAGS`/`-Dc_args` being set), which means `-O0`, so it may be worth tweaking them as well (after we see the results here). 

> @mrc0mmand could you try to turn on the slow tests under ASan+UBsan while you're at it to see how they fare with `-O1`?

Nice catch, would completely forgot about them :-)

mrc0mmand:
Note to myself: I need to backport several patches to RHEL 8, so `-O1` and higher don't report uninitialized variables.

keszybz:
LGTM.

mrc0mmand:
@evverx, @keszybz: now the build fails with:

```
 archlinux_systemd_sanitizers: FAILED: systemd-socket-activate@exe/src_activate_activate.c.o 
    archlinux_systemd_sanitizers: cc -Isystemd-socket-activate@exe -I. -I.. -Isrc/basic -I../src/basic -Isrc/boot -I../src/boot -Isrc/shared -I../src/shared -Isrc/systemd -I../src/systemd -Isrc/journal -I../src/journal -Isrc/journal-remote -I../src/journal-remote -Isrc/nspawn -I../src/nspawn -Isrc/resolve -I../src/resolve -Isrc/timesync -I../src/timesync -I../src/time-wait-sync -Isrc/login -I../src/login -Isrc/udev -I../src/udev -Isrc/libudev -I../src/libudev -Isrc/core -I../src/core -Isrc/shutdown -I../src/shutdown -I../src/libsystemd/sd-bus -I../src/libsystemd/sd-device -I../src/libsystemd/sd-event -I../src/libsystemd/sd-hwdb -I../src/libsystemd/sd-id128 -I../src/libsystemd/sd-netlink -I../src/libsystemd/sd-network -I../src/libsystemd/sd-resolve -Isrc/libsystemd-network -I../src/libsystemd-network -I../ -fdiagnostics-color=always -fsanitize=address,undefined -fno-omit-frame-pointer -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -std=gnu99 -g -Wextra -Werror=undef -Wlogical-op -Wmissing-include-dirs -Wold-style-definition -Wpointer-arith -Winit-self -Wfloat-equal -Wsuggest-attribute=noreturn -Werror=missing-prototypes -Werror=implicit-function-declaration -Werror=missing-declarations -Werror=return-type -Werror=incompatible-pointer-types -Werror=format=2 -Wstrict-prototypes -Wredundant-decls -Wmissing-noreturn -Wimplicit-fallthrough=5 -Wshadow -Wendif-labels -Wstrict-aliasing=2 -Wwrite-strings -Werror=overflow -Werror=shift-count-overflow -Werror=shift-overflow=2 -Wdate-time -Wnested-externs -Wno-unused-parameter -Wno-missing-field-initializers -Wno-unused-result -Wno-format-signedness -Wno-error=nonnull -Wno-maybe-uninitialized -ffast-math -fno-common -fdiagnostics-show-option -fno-strict-aliasing -fvisibility=hidden -fstack-protector -fstack-protector-strong --param=ssp-buffer-size=4 -Werror=shadow -include config.h -g -O1 -fno-omit-frame-pointer -ftrapv -pthread -MD -MQ 'systemd-socket-activate@exe/src_activate_activate.c.o' -MF 'systemd-socket-activate@exe/src_activate_activate.c.o.d' -o 'systemd-socket-activate@exe/src_activate_activate.c.o' -c ../src/activate/activate.c
    archlinux_systemd_sanitizers: ../src/activate/activate.c: In function ‘main’:
    archlinux_systemd_sanitizers: cc1: error: argument 1 range [18446744065119617112, 18446744065119617128] exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
    archlinux_systemd_sanitizers: cc1: note: in a call to built-in allocation function ‘__builtin_alloca_with_align’
    archlinux_systemd_sanitizers: cc1: all warnings being treated as errors
```

I've seen it only one (in a VM as well) and thought it was caused by misconfirugartion. What's the best way to fix/workaround it?

mrc0mmand:
Ugh, for some reason gcc 8.3.x doesn't like the `arg_listen` variable in `parse_args()`, because when I do:

```diff
diff --git a/src/activate/activate.c b/src/activate/activate.c
index 15a3150666..e374f6b723 100644
--- a/src/activate/activate.c
+++ b/src/activate/activate.c
@@ -386,7 +386,7 @@ static int parse_argv(int argc, char *argv[]) {
                         return version();

                 case 'l':
-                        r = strv_extend(&arg_listen, optarg);
+                        r = strv_extend(&arg_args, optarg);
                         if (r < 0)
                                 return log_oom();
```

The compilation finishes successfully (just ignore the fact that the patch makes no sense). After undoing the change, we're back with the error above. As it happens only with gcc 8.3.x, it almost looks like a bug in gcc.

mrc0mmand:
Nevermind, I think I managed to find the root cause, will file a issue/PR shortly.

keszybz:
Merge?

mrc0mmand:
I'd say we can merge this, as the last thing failing were the slow tests, which I dropped in the last force push.

PR #145, opened 2019-07-17
Performance improvements

mrc0mmand:
@evverx 

Following the proposal in https://github.com/systemd/systemd/pull/13087, I've put together a few changes which should help with the CentOS CI overload issue.

Changes:
1)  I moved the inline build scripts from Jenkins into this repo, which should improve transparency
2) The gcc part of the `systemd-pr-build-vagrant-sanitizers` job should be now disabled in favor of the clang one, to (hopefully temporarily) free up some resources
3) I proposed a *PR filter* to filter out some out-of-scope PRs from being tested

The PR filter limits the scope of changes to these folders/files:
```
$ SCOPE_RX='(^(catalog|factory|hwdb|meson.*|network|[^\.].*\.d$|rules|src|test|units))'
$ ls -a | grep -E "$SCOPE_RX"
catalog
factory
hwdb
meson.build
meson_options.txt
modprobe.d
network
rules
src
sysctl.d
sysusers.d
test
tmpfiles.d
units

```

However, to test the man-page generation, I relaxed the regex in the CentOS 7 job (`systemd-pr-build`) so we have it tested at least somewhere. If I managed to omit something from the regex, let me know.

Hopefully, this and the changes from #144 will make the CI a bit faster. (Right now the load statistics graph looks like something made by Bob Ross:
image
)

### DIFF
--- a/jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
+++ b/jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
@@ -24,5 +24,5 @@ fi
 git clone https://github.com/systemd/systemd-centos-ci
 cd systemd-centos-ci
 
-./agent-control.py --no-index --vagrant arch-sanitizers-gcc $ARGS
+#./agent-control.py --no-index --vagrant arch-sanitizers-gcc $ARGS
 ./agent-control.py --no-index --vagrant arch-sanitizers-clang $ARGS
--- a/jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
+++ b/jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+set -e
+
+function at_exit() {
+    # Correctly collect artifacts from all sanitizer jobs and generate a nice
+    # directory structure
+    mkdir _artifacts_all
+    mv artifacts_* _artifacts_all
+    mv _artifacts_all artifacts_all
+
+    utils/generate-index.sh artifacts_all index.html
+}
+
+trap at_exit EXIT
+
+export PATH="/home/systemd/bin:$PATH"
+ARGS=
+
+if [ "$ghprbPullId" ]; then
+    ARGS="$ARGS --pr $ghprbPullId "
+fi
+
+git clone https://github.com/systemd/systemd-centos-ci
+cd systemd-centos-ci
+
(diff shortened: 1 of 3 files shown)


jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
```diff
@@ -0,0 +1,28 @@
...
+
+export PATH="/home/systemd/bin:$PATH"
+ARGS=
+
+if [ "$ghprbPullId" ]; then
+    ARGS="$ARGS --pr $ghprbPullId "
+fi
+
+git clone https://github.com/systemd/systemd-centos-ci
+cd systemd-centos-ci
```
keszybz: I guess this makes things slower... but transparency is more important.

jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
```diff
@@ -19,6 +20,17 @@ ARGS=
...
+    # We're not testing the master branch, so let's see if the PR scope
+    # is something we should indeed test
+    git clone https://github.com/systemd/systemd systemd-tmp && cd systemd-tmp
+    git fetch -fu origin "refs/pull/$ghprbPullId/head:pr"
+    git checkout pr
+    git rebase master
+    if ! git diff master --name-only | grep -E '(^(catalog|factory|hwdb|meson*|network|[^\.].*.d|rules|src|test|units))'; then
+        echo "Changes in this PR don't seem relevant, skipping..."
+        exit 0
+    fi
```
keszybz: Is the rebase necessary? Maybe just do `git diff $(git merge-base master) instead`?

jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
mrc0mmand: This was, unfortunately, even in the original inline script, because the workspace is the systemd repo *context*, when triggered via systemd PR (i.e. when a job is triggered via PR event from the systemd repository, the root of the workspace is actually root of the systemd repository checked out to a given PR).

jenkins/runners/systemd-pr-build-vagrant-sanitizers.sh
mrc0mmand: Ooh, this is neat, thank you, didn't know about this!

mrc0mmand:
I'll go ahead and merge this one as well, because it won't have any effect until I re-configure the Jenkins jobs.

PR #146, opened 2019-07-17
Revert Arch Linux repo pin

mrc0mmand:
Let's see how systemd fares with gcc 9.1.x once again, after the bunch of fixes & workaround.

PR #148, opened 2019-07-18
vagrant: reload dbus policies before running systemd-networkd-tests.py

yuwata:


yuwata:
```
Got message type=error sender=org.freedesktop.DBus destination=:1.728 path=n/a interface=n/a member=n/a cookie=3 reply_cookie=3 signature=s error-name=org.freedesktop.DBus.Error.AccessDenied error-message=Rejected send message, 2 matched rules; type="method_call", sender=":1.728" (uid=0 pid=6256 comm="/build/build/resolvectl dns dummy98 10.10.10.12 10") interface="org.freedesktop.network1.Manager" member="SetLinkDNS" error name="(unset)" requested_reply="0" destination="org.freedesktop.network1" (uid=981 pid=6233 comm="/build/build/systemd-networkd ")
```

mrc0mmand:
@yuwata all Vagrant jobs should now use the latest gcc/kernel once again (since #146 was merged), so I dropped the workarounds for that from your PR.

yuwata:
@mrc0mmand Thanks.

yuwata:
@mrc0mmand I found the root of the issue. With sanitizers, the dbus config file is not the one provided by the current git master. So, systemd-networkd-tests.py cannot use new bus methods. Please make sure that the file `/usr/share/dbus-1/system.d/org.freedesktop.network1.conf` is provided by the build directory.

mrc0mmand:
> @mrc0mmand I found the root of the issue. With sanitizers, the dbus config file is not the one provided by the current git master. So, systemd-networkd-tests.py cannot use new bus methods. Please make sure that the file `/usr/share/dbus-1/system.d/org.freedesktop.network1.conf` is provided by the build directory.

Thanks a lot for the investigation! That actually makes sense, as we intentionally skip the `ninja -C build install` step in sanitizer runs to improve stability of the SUT. Will definitely look into it to find a reasonable compromise.

mrc0mmand:
@yuwata I pushed a potential fix. It's not particularly nice, but I'm not sure if there's a way to tell meson which files to install, instead of installing everything (at least I couldn't find any).

yuwata:
> we intentionally skip the `ninja -C build install` step in sanitizer runs to improve stability of the SUT.

I see. Thats's why `test_udev_driver`, which was dropped by https://github.com/systemd/systemd/pull/12984, failed in the sanitizer CI. I will re-enable the test.

PR #150, opened 2019-07-21
Improve the systemd-networkd testsuite output

mrc0mmand:


mrc0mmand:
Note for my possible future self: the testsuite output could be improved by using the [HTMLTestRunner](https://pypi.org/project/html-testRunner/) (mentioned [here](https://www.dev2qa.com/python-3-unittest-html-and-xml-report-example/))

PR #153, opened 2019-07-24
vagrant: collect output from a serial console

mrc0mmand:


mrc0mmand:
Note: this fail might need some investigation later

PR #155, opened 2019-07-30
agent/vagrant: run test/networkd-test.py

mrc0mmand:


mrc0mmand:
@yuwata hmm, is this something "new" or it is caused by `test/networkd-test.py` being run before `test/systemd-networkd-testsuite.sh`?

```
======================================================================
FAIL: test_udev_link_file (__main__.NetworkctlTests)
----------------------------------------------------------------------
  File "./test/test-network/systemd-networkd-tests.py", line 450, in test_udev_link_file
    self.assertRegex(output, r'Link File: (?:/usr)/lib/systemd/network/99-default.link')
AssertionError: Regex didn't match: 'Link File: (?:/usr)/lib/systemd/network/99-default.link' not found in '● 115: test1                                                 \n             Link File: n/a                                  \n          Network File: /run/systemd/network/11-dummy.network\n                  Type: ether                                \n                 State: degraded (configured)                \n                Driver: dummy                                \n            HW Address: ca:78:bf:ff:64:3b                    \n                   MTU: 1500                                 \n  Queue Length (Tx/Rx): 1/1                                  \n               Address: fe80::c878:bfff:feff:643b'
```

yuwata:
Hmm, I am not sure why it fails in this case. But I think the test `test_udev_link_file` also needs to be dropped, as it uses `-w` option of `udevadm`.

PR #157, opened 2019-08-01
agent: optimize RHEL7/RHEL8 builds as well

mrc0mmand:


mrc0mmand:
Unfortunately, the upstream diverged too far from the RHEL8 branch, so a patch, which would fix all warnings, would basically end up being a rebase.

PR #161, opened 2019-08-12
Collect potential coredumps after running the testsuite 

### LINKED ISSUE #160: Check for possible coredumps after running the testsuite
Let's incorporate `coredumpctl --since "..."` and `coredumpctl info --since "..."` into the CI to check if certain systemd subcomponents don't *silently* crash during the testing process. 

This should allow us to catch issues like https://github.com/systemd/systemd/pull/12957#issuecomment-520227344

mrc0mmand:
Fixes #160 

---

~~So far this shows only stack traces from all coredumps. Calling `bt full` on every coredump would be definitely handy, but right now this would be somewhat cumbersome, as `coredumpctl` doesn't allow overriding/editing the `$SYSTEMD_DEBUGGER` command line, and `.gdbinit` runs before the gdb is properly initialized with passed arguments (executable & coredump).~~

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -42,7 +42,7 @@ sed -i 's/skip_if_unavailable=True/skip_if_unavailable=False/' "$COPR_REPO_PATH"
 if ! rpm --import https://copr-be.cloud.fedoraproject.org/results/mrc0mmand/systemd-centos-ci/pubkey.gpg; then
     rpm --import http://artifacts.ci.centos.org/systemd/mrc0mmand-systemd-centos-ci/pubkey.gpg
 fi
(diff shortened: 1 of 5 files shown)


mrc0mmand:
Example artifacts from a *failing* build: https://ci.centos.org/job/systemd-ci-build/775/artifact//index.html

mrc0mmand:
TODO: run `coredumpctl_collect` on journals from `TEST-??-*` under QEMU (the nspawn variations are already covered by the system journal)

mrc0mmand:
Artifacts for systemd/systemd#12957: https://ci.centos.org/job/systemd-ci-build/783/artifact//index.html

PR #165, opened 2019-08-22
vagrant: collect coredumps using coredumpctl in sanitizer runs

### LINKED ISSUE #164: Extend the coredumpctl stuff from #161 to the sanitizer run as well
Collecting coredumps as done in #161 should be useful in the sanitizer run as well.

mrc0mmand:
Continuation of #161

Fixes #164


mrc0mmand:
TODO:

- fix the commit message for f0024cb (s/split/drop/)
- skip the `clean-again` phase for `TEST-01-BASIC`, so we can keep the journal for later coredump analysis (delete it only if the coredumpctl check passes)

PR #168, opened 2019-09-26
vagrant: enable nested KVM

mrc0mmand:
As nested KVM got much more stable in recent years, let's give it a try
and attempt to use it by default to speed up tests under Vagrant.

The nested KVM setup requires two steps:
1) Enable nested KVM on the host system (modprobe kvm_* nested=1)
2) Tell the vagrant-libvirt plugin to use it (domain.nested = true)

### DIFF
--- a/vagrant/vagrant-setup.sh
+++ b/vagrant/vagrant-setup.sh
@@ -6,8 +6,32 @@
 
 VAGRANT_PKG_URL="https://releases.hashicorp.com/vagrant/2.2.4/vagrant_2.2.4_x86_64.rpm"
 
-set -e -u
 set -o pipefail
+set -e -u
+
+# Set up nested KVM
+# Let's make all errors "soft", at least for now, as we're still perfectly
+# fine with running tests without nested KVM
+if KVM_MODULE_NAME="$(lsmod | grep -m1 -Eo '(kvm_intel|kvm_amd)')"; then
+    echo "[vagrant-setup] Detected KVM module: $KVM_MODULE_NAME"
+    # Attempt to reload the detected KVM module with nested=1 parameter
+    if modprobe -v -r $KVM_MODULE_NAME && modprobe -v $KVM_MODULE_NAME nested=1; then
+        # The reload was successful, check if the module's 'nested' parameter
+        # confirms that nested KVM is indeed enabled
+        KVM_MODULE_NESTED="$(< /sys/module/$KVM_MODULE_NAME/parameters/nested)" || :
+        echo "[vagrant-setup] /sys/module/$KVM_MODULE_NAME/parameters/nested: $KVM_MODULE_NESTED"
+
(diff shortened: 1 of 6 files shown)


vagrant/vagrant-setup.sh
```diff
@@ -6,8 +6,29 @@
 
 VAGRANT_PKG_URL="https://releases.hashicorp.com/vagrant/2.2.4/vagrant_2.2.4_x86_64.rpm"
 
-set -e -u
 set -o pipefail
+set -u
+
+# Set up nested KVM
+KVM_MODULE_NAME="$(lsmod | grep -m1 -Eo '(kvm_intel|kvm_amd)')"
```
filbranden: I suggest keeping the `set -e` active.

Just use a `|| :` at the end of this line to prevent the error from here:

```sh
# Set up nested KVM, if already running inside KVM.
KVM_MODULE_NAME="$(lsmod | grep -m1 -Eo '(kvm_intel|kvm_amd)')" || :
```

You can do the same if there are other lines that might trigger an error due to non-zero exit that's expected.

vagrant/vagrant-test.sh
```diff
@@ -21,6 +21,10 @@ fi
 
 pushd /build || { echo >&2 "Can't pushd to /build"; exit 1; }
 
+# Trick the QEMU-related test functions, so it appears we're not already
+# running under KVM
+sed -i 's/systemd-detect-virt -v/echo not-kvm/g' test/test-functions
```
filbranden: I imagine this goes away once systemd/systemd#13666 is merged, right?

vagrant/vagrant-test.sh
```diff
@@ -103,7 +107,7 @@ done
 ## Other integration tests ##
 TEST_LIST=(
     "test/test-exec-deserialization.py"
-    "test/test-network/systemd-networkd-tests.py"
+#    "test/test-network/systemd-networkd-tests.py"
```
filbranden: Unsure why this one needed to be commented out... I'm guessing that's also temporary, right?

vagrant/vagrant-test.sh
mrc0mmand: Yes, that's only for debugging.

vagrant/vagrant-test.sh
mrc0mmand: And this one too. I just wanted to keep the resources busy for as short period of time as possible, as the nested KVM stuff is related only to the integration tests.

vagrant/vagrant-setup.sh
mrc0mmand: The `set -e` is still active, I simply moved it just after this function. The idea here is that the nested KVM stuff is not *crucial* for the testing itself - it can run fine without it. I agree that having it enable from the start it way nicer though. I'll try to rewrite it a little bit.

PR #169, opened 2019-10-07
agent: pin meson to v0.51.2

mrc0mmand:
The latest meson version (0.52.0) introduced an issue, where the
configure/setup phase ends up in an endless loop. Until it's resolved,
let's pin meson to its last working version.

See: systemd/systemd#13742


mrc0mmand:
The *sanity* part of the CI passed, merging...

PR #194, opened 2019-11-21
agent: bump the upstream CentOS 7 job to CentOS 8

mrc0mmand:
This will help with several things, especially:
 - drop the plethora of CentOS 7 vs upstream systemd workarounds, like
   the custom Copr repo with various *mangled* dependencies
 - support systemd-homed (see: systemd/systemd#14096)
 - run more tests which depend on newer kernel's features

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -117,6 +117,15 @@ grubby --args="user_namespace.enable=1" --update-kernel="$(grubby --default-kern
 grep -r "user_namespace.enable=1" /boot/loader/entries/
 echo "user.max_user_namespaces=10000" >> /etc/sysctl.conf
 
+# Following steps are needed to 'unconfuse' systemd after being replaced by
+# an upstream version
+# 1) user-0.slice get stuck for a while, which breaks ssh connections
+# 2) systemd-reboot.service has an incompatible format, so daemon-reexec
+#    is needed to fix this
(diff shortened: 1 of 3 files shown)


evverx:
@mrc0mmand I think in terms of covering older kernels to make sure `systemd` can be compiled and run there without major issues it would be better to keep CentOS 7. Newer kernels are already tested on Ubuntu CI, Travis CI and CentOS CI (Arch in KVM). Probably it would make sense `systemd-homed` optional and turn it off completely when it can't be supported for whatever reasons.

mrc0mmand:
> @mrc0mmand I think in terms of covering older kernels to make sure `systemd` can be compiled and run there without major issues it would be better to keep CentOS 7. Newer kernels are already tested on Ubuntu CI, Travis CI and CentOS CI (Arch in KVM). Probably it would make sense to make `systemd-homed` optional and turn it off completely when it can't be supported for whatever reasons.

That's the kind of feedback I was looking for, thanks! :-)

I wasn't sure if *ancient* kernels weren't already tested by Ubuntu CI, but apparently even Xenial has newer kernel than CentOS 7 (relatively speaking). In that case, it would indeed make sense to keep the CentOS 7 job alive and just disable `systemd-homed`. I'll do it in a separate PR and keep this one as a reference, should we get our hands on additional resources to be able to run both CentOS jobs simultaneously.

PR #202, opened 2019-12-16
Use -Dinstall-tests=true in more places

keszybz:
This is for https://github.com/systemd/systemd/pull/14338.
Now all TEST-* tests need builds with -Dinstall-tests=true.

### DIFF
--- a/vagrant/bootstrap_scripts/arch-sanitizers-clang.sh
+++ b/vagrant/bootstrap_scripts/arch-sanitizers-clang.sh
@@ -35,6 +35,7 @@ meson build \
       -Dtests=unsafe \
       -Ddbuspolicydir=/usr/share/dbus-1/system.d \
       -Dman=false \
+      -Dinstall-tests=true \
       -Db_sanitize=address,undefined \
       -Db_lundef=false # See https://github.com/mesonbuild/meson/issues/764
 ninja -C build
--- a/vagrant/bootstrap_scripts/arch-sanitizers-gcc.sh
+++ b/vagrant/bootstrap_scripts/arch-sanitizers-gcc.sh
@@ -30,6 +30,7 @@ meson build \
       -Dtests=unsafe \
       -Ddbuspolicydir=/usr/share/dbus-1/system.d \
       -Dman=false \
+      -Dinstall-tests=true \
       -Db_sanitize=address,undefined
 ninja -C build
 


mrc0mmand:
Thanks! One concern though: I dropped `-Dinstall-tests=true` in https://github.com/systemd/systemd-centos-ci/commit/f3d6fcb76444237ccbdd286b40d176eede71fc65 in sanitizer jobs due to https://github.com/systemd/systemd/pull/13886#issuecomment-548830867 - won't this cause issues once again?

keszybz:
Oh, so this is an exact revert of your commit. I think that your commit looks wrong:
> However, in case of installed tests (-Dinstall-tests=true), these tests
are under /lib/systemd/tests, but the *correct* libsystemd DSO is not in
the binary's RPATH (as it's installed usually under /lib64), which
results in the system one being used. And as we don't run `ninja -C build
install`, the system libsystemd DSO has a different version than the
built one, which leads to nasty errors.

We can't meaningfully run tests (or other binaries in fact) from temporary install prefix.

mrc0mmand:
> Oh, so this is an exact revert of your commit. I think that your commit looks wrong:
> 
> > However, in case of installed tests (-Dinstall-tests=true), these tests
> > are under /lib/systemd/tests, but the _correct_ libsystemd DSO is not in
> > the binary's RPATH (as it's installed usually under /lib64), which
> > results in the system one being used. And as we don't run `ninja -C build install`, the system libsystemd DSO has a different version than the
> > built one, which leads to nasty errors.
> 
> We can't meaningfully run tests (or other binaries in fact) from temporary install prefix.

Ah, I see. Just to make things clear - I thought that `-Dinstall-tests=true` is relevant only for `TEST-24-UNIT-TESTS` which makes use of tests under `/lib/systemd/tests`, thus for `meson test` and other integration tests the option is somewhat pointless. Has this fact changed?

keszybz:
Yes. #14338 makes it required to have `-Dinstall-tests=true` for all the functional tests.

mrc0mmand:
Hm, in that case I'm not sure how to proceed here. We can't run `ninja -C build install` on the machine, as it would make the underlying system pretty unstable (and slow), because of the sanitizers. In the sanitizers run we run only `TEST-01-BASIC` under qemu & nspawn to make at least one full boot of systemd with sanitizers, but that's it, so we don't really need installed tests here. Would it be possible to switch this requirement off via some env variable?

/cc @evverx

keszybz:
I will try to partially revert the patch to make TEST-01-BASIC do manual install again.

mrc0mmand:
Closing as this was addressed in https://github.com/systemd/systemd/pull/14338 and we can run `TEST-01-BASIC` again without installing the just built systemd revision, thanks again for the rewrite effort!

(Also, I just noticed I already did the exact same change in https://github.com/systemd/systemd-centos-ci/commit/73d4c6067d8b8983ca298c0004544b8f4ab85b36, *sigh*)

PR #217, opened 2020-01-30
vagrant: run certain other integration tests under sanitizers

mrc0mmand:
Let's slightly extend the number of integration tests (`TEST-??-*`) we run under sanitizers to cover some of the larger systemd subcomponents introduced recently:

1) `TEST-04-JOURNAL` to cover `systemd-journald`
2) `TEST-45-REPART` to cover `systemd-repart` 
3) `TEST-46-HOMED` to cover `systemd-homed` and related stuff

~~`TEST-45-REPART` needs a small tweak to work correctly under sanitizers because the used `grep` statements were tripping over LSan which then killed `grep` during post-processing. Simple `LD_PRELOAD=` in front of each `grep` does the trick.~~ Fixed by systemd/systemd#14715

Also, I finally disabled `test-execute/exec-specifier-interpolation` as the completely random LSan crashes were driving me insane and I couldn't reproduce them on both Arch Linux and Fedora Rawhide after letting it loop for several hours straight.

/cc @evverx

### DIFF
--- a/vagrant/vagrant-test-sanitizers.sh
+++ b/vagrant/vagrant-test-sanitizers.sh
@@ -41,8 +41,30 @@ fi
 # test-journal-flush: unstable on nested KVM
 echo 'int main(void) { return 77; }' > src/journal/test-journal-flush.c
 
+## Temporary wrapper for `meson test` which disables LSan for `test-execute`
+# LSan keeps randomly crashing during `test-execute` so let's (temporarily)
+# disable it until we find out the culprit.
+# See:
+#   https://github.com/systemd/systemd-centos-ci/pull/217#issuecomment-580717687
+#   https://github.com/systemd/systemd/issues/14598
+ASAN_WRAPPER="$(mktemp $PWD/build/asan-wrapper-XXX.sh)"
+cat > "$ASAN_WRAPPER" << EOF
+#!/bin/bash
+
+export ASAN_OPTIONS=$ASAN_OPTIONS
+export UBSAN_OPTIONS=$UBSAN_OPTIONS
+
+if [[ \$(basename "\$1") == 'test-execute' ]]; then
+    ASAN_OPTIONS="\$ASAN_OPTIONS:detect_leaks=0"
+fi
+
+exec "\$@"
+EOF
+
+chmod +x "$ASAN_WRAPPER"
+
 # Run the internal unit tests (make check)
-exectask "ninja-test_sanitizers" "meson test -C build --print-errorlogs --timeout-multiplier=3"
+exectask "ninja-test_sanitizers" "meson test -C build --wrapper=$ASAN_WRAPPER --print-errorlogs --timeout-multiplier=3"
 
 ## Run TEST-01-BASIC under sanitizers
 # Set timeouts for QEMU and nspawn tests to kill them in case they get stuck
--- a/vagrant/vagrant-test-sanitizers.sh
+++ b/vagrant/vagrant-test-sanitizers.sh
@@ -8,16 +8,23 @@
 
 DISTRO="${1:-unspecified}"
 SCRIPT_DIR="$(dirname $0)"
-# task-control.sh is copied from the systemd-centos-ci/common directory by vagrant-builder.sh
+# Following scripts are copied from the systemd-centos-ci/common directory by vagrant-builder.sh
 . "$SCRIPT_DIR/task-control.sh" "vagrant-$DISTRO-testsuite" || exit 1
 . "$SCRIPT_DIR/utils.sh" || exit 1
 
 pushd /build || { echo >&2 "Can't pushd to /build"; exit 1; }
 
-# Sanitizer-specific options
+## Sanitizer-specific options
 export ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
 export UBSAN_OPTIONS=print_stacktrace=1:print_summary=1:halt_on_error=1
 
+## To be able to run integration tests under sanitizers we have to use the dynamic
+## versions of sanitizer libraries, especially when it comes to ASAn. With gcc
+## it's quite easy as ASan is compiled dynamically by default there and all necessary
+## libraries are in standard locations.
+## With clang things get a little bit complicated as we need to explicitly tell clang
+## to use the dynamic ASan library and then instruct the rest of the system
+## to where it can find it, as it is in a non-standard library location.
 _clang_asan_rt_name="$(ldd build/systemd | awk '/libclang_rt.asan/ {print $1; exit}')"
 
 if [[ -n "$_clang_asan_rt_name" ]]; then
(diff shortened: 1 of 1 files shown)


mrc0mmand:
TODO:
 - add `TEST-04-JOURNAL` as well
 - try to introduce a memory leak
 - drop the debug-only commit

evverx:
It seems `test-execute` keeps failing under LSan. I think at this point it would make sense to bring back the tests we turned off and turn off LSan by passing ASAN_OPTIONS=detect_leaks=0 when `test-execute` is run at least until we figure out what it is that keeps crashing it.

mrc0mmand:
> It seems `test-execute` keeps failing under LSan.

Hmpf, interesting, so it's not fault of the `exec-specifier-interpolation` as I originally suspected. 

>  I think at this point it would make sense to bring back the tests we turned off and turn off LSan by passing ASAN_OPTIONS=detect_leaks=0 when `test-execute` is run at least until we figure out what it is that keeps crashing it.

This may be a little bit tricky, if we want to set it *only* for `test-execute`. Not sure if there's a nicer way, but I managed to ~~fix~~ do that by patching the original `meson.build` file:

```diff
diff --git a/meson.build b/meson.build
index ea69276019..37b1389d06 100644
--- a/meson.build
+++ b/meson.build
@@ -3015,9 +3015,13 @@ foreach tuple : tests
                 elif type == 'unsafe' and want_tests != 'unsafe'
                         message('@0@ is an unsafe test'.format(name))
                 elif want_tests != 'false'
-                        test(name, exe,
-                             env : test_env,
-                             timeout : timeout)
+                        if name == 'test-execute'
+                                temp_env = test_env
+                                temp_env.set('ASAN_OPTIONS', 'strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1:detect_leaks=0')
+                                test(name, exe, env : temp_env, timeout : timeout)
+                        else
+                                test(name, exe, env : test_env, timeout : timeout)
+                        endif
                 endif
         else
                 message('Not compiling @0@ because @1@ is not true'.format(name, condition))

```

mrc0mmand:
Okay, this *might* be much nicer and easier to do:

```diff
diff --git a/src/test/test-execute.c b/src/test/test-execute.c
index ff40e0dd43..e9a86b2d2f 100644
--- a/src/test/test-execute.c
+++ b/src/test/test-execute.c
@@ -859,6 +859,8 @@ int main(int argc, char *argv[]) {
                 log_notice("Running on TravisCI under ASan, skipping, see https://github.com/systemd/systemd/issues/10696");
                 return EXIT_TEST_SKIP;
         }
+
+        putenv((char*)"ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1:detect_leaks=0");
 #endif
 
         (void) unsetenv("USER");

```

@evverx I guess having this only in CentOS CI would make sense as I haven't seen the issue anywhere else.

evverx:
I think it goes sideways somewhere along the way. `exec-specifier-interpolation` just happens to be the last test as far as I can tell :-)

> @evverx I guess having this only in CentOS CI would make sense as I haven't seen the issue anywhere else.

Agreed.

Regarding detect_leaks=0, in theory it should be possible to write a bash script that would look like
```sh
#!/bin/bash

export ASAN_OPTIONS=...
export UBSAN_OPTIONS=...

if [[ $(basename "$1") == 'test-execute' ]]; then
    ASAN_OPTIONS="$ASAN_OPTIONS:detect_leaks=0" exec "$1"
else
    exec "$1"
fi
```
and then pass it to `meson test` via the "wrapper" option.

mrc0mmand:
That's definitely better than `sed`-ing source files, thanks! :-)

After replacing `$1` with `$@` it seems to work (locally) as expected (as the `check-compilation/check-help` tests have multiple arguments), let's see how it fares in CI.

evverx:
> After replacing $1 with $@ it seems to work

I just went where the flow took me and somehow wrote `$1` instead of `$@` :-) Thanks for spotting it!

Not that it matters, but having looked at the script a little bit closer, I'd also drop the "else" part and move `exec "$@"` after the "if" statement:
```sh
if [[ $(basename "$1") == 'test-execute' ]]; then
    ASAN_OPTIONS="$ASAN_OPTIONS:detect_leaks=0"
fi

exec "$@"
```

evverx:
@mrc0mmand I think to make sure it works it would be easier to temporarily revert https://github.com/systemd/systemd-centos-ci/pull/215 to see whether `TEST-*-REPART` is able to catch the double-free (assuming `util-linux` hasn't been released yet).

By the way, could you switch back to `clang`? Looks like `gcc` didn't help much anyway :-)

mrc0mmand:
> @mrc0mmand I think to make sure it works it would be easier to temporarily revert #215 to see whether `TEST-*-REPART` is able to catch the double-free (assuming `util-linux` hasn't been released yet).


Good point, didn't think of that :-)

> 
> By the way, could you switch back to `clang`? Looks like `gcc` didn't help much anyway :-)

Done!

mrc0mmand:
I forgot that a direct revert of #215 won't work as this PR doesn't rebuilt the underlying CI image, so let's just explicitly upgrade the offending packages during the bootstrap phase.

mrc0mmand:
Looking at the results it looks like it's working as intended https://ci.centos.org/job/systemd-ci-build/1043/artifact//artifacts_all/artifacts_sM91Bq/vagrant-logs.KNg/vagrant-arch-sanitizers-clang-testsuite.qif/TEST-45-REPART_sanitizer_errors_FAIL.log

mrc0mmand:
Surprisingly enough, my memory leak attempt was not caught by the CI even though when I run the same test locally, it fails and complains about leaked memory:

```
...
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]: =================================================================
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]: ==63==ERROR: LeakSanitizer: detected memory leaks
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]: Direct leak of 16 byte(s) in 1 object(s) allocated from:
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]:     #0 0x7fe652025e2f in malloc (/usr/lib64/clang/9.0.0/lib/libclang_rt.asan-x86_64.so+0xeae2f)
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]:     #1 0x491945 in run /home/mrc0mmand/repos/systemd/build/../src/home/homed.c:16:11
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]:     #2 0x49163d in main /home/mrc0mmand/repos/systemd/build/../src/home/homed.c:50:1
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]:     #3 0x7fe65053a1a2 in __libc_start_main (/lib64/libc.so.6+0x271a2)
Jan 31 19:10:20 systemd-testsuite systemd-homed[63]: SUMMARY: AddressSanitizer: 16 byte(s) leaked in 1 allocation(s).
```

Will need to take a closer look at that.

mrc0mmand:
Nevermind, the leak was in a particularly dumb spot and when run manually in CentOS CI on Arch the `systemd-homed` binary won't show anything either (as opposed to my machine) unless run under an unprivileged user. Moving the leak to a better location showed that it works as expected.

Issue #216, opened 2020-01-28
vagrant: revert #215 once a new util-linux version is released & packaged in Arch Linux


mrc0mmand:
See: #215

PR #219, opened 2020-02-03
vagrant: unpin util-linux

mrc0mmand:
A new version of util-linux ([2.35.1](https://github.com/karelzak/util-linux/releases/tag/v2.35.1)) has been released which should address the double-free issue. ~~Let's check that before reverting #215 (and thus fixing #216).~~ Seems to work as expected, let's revert the workaround.


mrc0mmand:
The networkd testsuite issue is unrelated and currently being worked on.

PR #229, opened 2020-03-23
agent/vagrant: make the parallelization temporarily conditioned

mrc0mmand:
The test suite is being rewritten in systemd/systemd#14338 which should
speed things up significantly. However, it also breaks the current
parallelization optimizations done by test runners in CI, so let's make
them temporarily conditional for branches which contain the new code.


keszybz:
Thanks! Let's merge this, and see how it goes.

PR #234, opened 2020-03-30
debug-only: parallelize after testsuite rewrite

mrc0mmand:


agent/testsuite.sh
```diff
@@ -104,11 +98,7 @@ for t in test/TEST-??-*; do
     rm -fr "$TESTDIR"
     mkdir -p "$TESTDIR"
 
-    if [[ $PARALLELIZE -ne 0 ]]; then
-        exectask_p "${t##*/}" "make -C $t clean setup run && touch $TESTDIR/pass"
-    else
-        exectask "${t##*/}" "make -C $t setup run && touch $TESTDIR/pass"
-    fi
+    exectask_p "${t##*/}" "make -C $t setup run && touch $TESTDIR/pass"
```
keszybz: I don't think this is going to work. Tests cannot be run from the same image in parallel. To do that, we'd need need some kind of ephemeral mode of running qemu.

mrc0mmand: Yeah... I was just curious what's going to break with parallelization and the answer is - a lot :-)

PR #235, opened 2020-03-31
agent/vagrant: tweak the journal location

mrc0mmand:
systemd/systemd#15281 introduces a change which collects all journals of
each integration tests and merges them into one located under
`$TESTDIR/system.journal`. Let's update the paths in CI scripts
accordingly.

---

I'll test this properly once systemd/systemd#15281 is merged.

### DIFF
--- a/agent/testsuite.sh
+++ b/agent/testsuite.sh
@@ -118,12 +118,12 @@ done
 # Save journals created by integration tests
 for t in test/TEST-??-*; do
     testdir="/var/tmp/systemd-test-${t##*/}"
-    if [[ -d "$testdir/journal" ]]; then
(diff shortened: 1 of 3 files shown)


mrc0mmand:
As for the `TEST-13-NSPAWN-SMOKE`, I've seen this fail several times already and the VM simply gives up and does nothing until it's killed by `timeout`. It can be seen [here](https://ci.centos.org/job/systemd-pr-build-vagrant/6332/artifact//systemd-centos-ci/index.html) as well (and with appropriate journals).

Not sure what's going on, but it looks like the level of abstraction here (bare -> KVM -> KVM -> nspawn) is just too much.

PR #242, opened 2020-04-14
agent/rhel8: attempt to collect coredumps from each SUT

mrc0mmand:


mrc0mmand:
Waiting for https://github.com/systemd-rhel/rhel-8/pull/85

PR #255, opened 2020-05-15
vagrant: use LDFLAGS="-Wl,--no-as-needed" in sanitizer runs

mrc0mmand:
Since version 10, both gcc and clang started to ignore certain linker errors
when compiling with -fsanitize=address. This eventually leads up to -lcrypt
not being correctly propagated, but the fact is masked by the aforementioned
issue. However, when the binary attempts to load a symbol from the libcrypt
binary, it crashes since it's not linked correctly against it.
Negating the -Wl,--as-needed used by default by -Wl,--no-as-needed seems to
help in this case.

See:
  https://bugzilla.redhat.com/show_bug.cgi?id=1827338#c3
  https://github.com/systemd/systemd-centos-ci/issues/247

/cc @evverx - this whole situation feels especially icky, hopefully I'm not missing something obvious :-)

### DIFF
--- a/jenkins/runners/systemd-cron-build.sh
+++ b/jenkins/runners/systemd-cron-build.sh
@@ -36,5 +36,5 @@ ARGS=
 git clone https://github.com/systemd/systemd-centos-ci
 cd systemd-centos-ci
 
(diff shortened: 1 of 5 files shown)


evverx:
As long as it works I think it should be good to go.

PR #258, opened 2020-05-19
agent: override RTC in early boot via systemd.clock_usec

mrc0mmand:


agent/bootstrap.sh
```diff
@@ -176,9 +176,29 @@ if ! lsinitrd -m /boot/initramfs-$(uname -r).img | grep "^systemd$"; then
...
-grubby --args="user_namespace.enable=1" --update-kernel="$(grubby --default-kernel)"
-grep "user_namespace.enable=1" /boot/grub2/grub.cfg
+GRUBBY_ARGS=(
+    # Needed for systemd-nspawn -U
+    "user_namespace.enable=1"
+    # As the RTC on CentOS CI machines is notoriously incorrect, let's override
+    # it early in the boot process to properly execute units using
+    # ConditionNeedsUpdate=
+    # See: https://github.com/systemd/systemd/issues/15724#issuecomment-628194867
+    "systemd.clock_usec=$(($(date +%s%N) / 1000))"
```
poettering, line 186: it probably would be more correct to round up here... otherwise the test machine might end up running with a clock slightly in the past, which we don't want.

Also, I hope that bash calculus is 64bit safe because these values are large

mrc0mmand: > it probably would be more correct to round up here... otherwise the test machine might end up running with a clock slightly in the past, which we don't want.

Good point, will do, thanks!
 
> Also, I hope that bash calculus is 64bit safe because these values are large

Thankfully yes (since v2.x iirc).

So far it's working as advertised:

````
May 19 23:50:06 n46.gusty.ci.centos.org kernel: rtc_cmos 00:01: setting system clock to 2020-05-19 23:25:27 UTC (1589930727)
...
May 19 23:50:06 n46.gusty.ci.centos.org systemd[1]: Set system clock to Tue 2020-05-19 21:50:05 UTC, as specified on the kernel command line.
```

PR #265, opened 2020-05-26
agent: use distro meson on CentOS 7

mrc0mmand:
to catch compatibility issues with older meson versions

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -45,10 +45,8 @@ fi
 yum -y install epel-release yum-utils gdb
 yum-config-manager --enable epel
 yum -y update
-yum -y install busybox dnsmasq e2fsprogs gcc-c++ libasan libbpf-devel libfdisk-devel nc net-tools ninja-build \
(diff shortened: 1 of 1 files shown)


evverx:
It seems `meson ... --optimization=g` below should be replaced with `CFLAGS=-Og CXXFLAGS=-Og meson ...` or with `-Dc_args=... -Dcpp_args` if those are supported there.

evverx:
Though I don't think whatever is passed via `c_args` is passed on to the fuzzers so it seems `CFLAGS` is the only option that works

mrc0mmand:
> It seems `meson ... --optimization=g` below should be replaced with `CFLAGS=-Og CXXFLAGS=-Og meson ...` or with `-Dc_args=... -Dcpp_args` if those are supported there.

Ah, thanks, missed this one!



> Though I don't think whatever is passed via `c_args` is passed on to the fuzzers so it seems `CFLAGS` is the only option that works

I guess this doesn't matter here, since CentOS 7 has a pretty old gcc without UBSan, so fuzzers are skipped there anyway.

mrc0mmand:
Also, as it failed (as expected) with ` ERROR: Tried to access unknown option "optimization".` I pointed it to https://github.com/systemd/systemd/pull/15937 to check if everything works as expected.

PR #278, opened 2020-06-09
agent/vagrant: don't remove the pre-initialized $TESTDIR

mrc0mmand:
After the test suite rewrite we do a test pre-initialization via
initialize_integration_tests() helper, so we shouldn't clean up the just
initialized $TESTDIR anymore.


mrc0mmand:
Fail in TEST-02-CRYPTSETUP is expected due to https://github.com/systemd/systemd/pull/16111.

PR #282, opened 2020-06-16
vagrant: work around current Vagrant's DSO hell

mrc0mmand:


mrc0mmand:
It works, but I really have to find a proper solution soon, until I gouge my eyes out...

PR #285, opened 2020-06-27
debug-only: consistent hwdb-update timeouts under gcc & sanitizers

mrc0mmand:
Since the latest hwdb update the hwdb-update service regularly timeouts
under gcc & sanitizers (which is currently run only in cron jobs). Let's
try to figure out why.


mrc0mmand:
Should be resolved by https://github.com/systemd/systemd/pull/16304.

PR #293, opened 2020-08-05
common: compress task logs

mrc0mmand:
TBD


mrc0mmand:
Oh well, Jenkins can't stream gzip-compressed files so the client (browser) can decompress & show them instead of downloading them, which would unnecessarily hinder the *user friendliness* - see https://issues.jenkins-ci.org/browse/JENKINS-45275. Let's close this, for now...

PR #294, opened 2020-08-20
vagrant: pin repositories to kernel 5.7.12 (2020-08-14)

mrc0mmand:
Since kernel 5.8.1 the test VMs started to randomly softlock when trying
to online non-boot CPUs. To make the CI stable again, let's revert back
to kernel 5.7.12.

See: https://bugs.archlinux.org/task/67649


mrc0mmand:
The CI fail is unrelated (but it's something I'll investigate further).

PR #308, opened 2020-11-11
Support alternative architectures (ppc64le & aarch64)

mrc0mmand:
Let's try to run tests on a sanitized build on newly supported alternative architectures.

TBD

---

[2020-12-09] There's no QEMU package for the ppc64le arch, and we can't run KVM there, since the machines are already virtualized, so the support for integration tests is not possible for now

### DIFF
--- a/agent/bootstrap-alt.sh
+++ b/agent/bootstrap-alt.sh
@@ -1,0 +1,211 @@
+#!/usr/bin/bash
+
+LIB_ROOT="$(dirname "$0")/../common"
+# shellcheck source=common/task-control.sh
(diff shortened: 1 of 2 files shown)


mrc0mmand:
Missing `util-linux` patches in RHEL 8:
 * https://github.com/karelzak/util-linux/commit/2f35c1ead621f42f32f7777232568cb03185b473 (libfdisk, breaks `test-repart`, would need to *tweak* the libfdisk dep in `meson.build` from `2.33` to `2.32.1` if backported)
 * [tbd] (libmount, memory leak)

karelzak:
@mrc0mmand add the issues to BZ and I'll try to backport it to RHEL-8.

PR #310, opened 2020-11-13
bootstrap: set -Dfuzz-tests=true

yuwata:
C.f. https://github.com/systemd/systemd/pull/17599.

### DIFF
--- a/agent/bootstrap-rhel8.sh
+++ b/agent/bootstrap-rhel8.sh
@@ -71,8 +71,7 @@ if setenforce 0; then
 fi
 
 # Compile systemd
-#   - slow-tests=true: enable slow tests => enables fuzzy tests using libasan
-#     installed above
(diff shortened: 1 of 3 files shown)


agent/bootstrap-rhel8.sh
```diff
@@ -129,6 +128,7 @@ fi
             -Ddefault-hierarchy=legacy
             # Custom options
             -Dslow-tests=true
+            -Dfuzz-tests=true
```
mrc0mmand, line 131: RHEL8 unfortunately doesn't support `-Dfuzz-tests=`, so please drop this chunk

mrc0mmand (changes requested):
Thanks for the patch! Just one thing to fix:

yuwata:
@mrc0mmand Thank you for the comment. Updated.

PR #336, opened 2021-01-28
Tweak the image initialization

mrc0mmand:
Since systemd/systemd@d9e606e each test use a base image which they
build on. However, with the current parallel test setup this introduces
a race condition, where each test tries to create the base image.

Let's initialize the base image separately and then run the rest of the
tests along with their setup phases.


keszybz:
This looks nice.

PR #344, opened 2021-02-23
vagrant: 'mesonize' the compiler flags

mrc0mmand:
Apparently, mixing CFLAGS/CXXFLAGS/LDFLAGS and their meson counterparts
is unsupported, so let's convert the existing flags to meson's
equivalents to keep the builds happy.

### DIFF
--- a/vagrant/bootstrap_scripts/arch-sanitizers-gcc.sh
+++ b/vagrant/bootstrap_scripts/arch-sanitizers-gcc.sh
@@ -37,7 +37,7 @@ rm -fr "$BUILD_DIR"
 # Compile systemd with the Address Sanitizer (ASan) and Undefined Behavior
 # Sanitizer (UBSan)
 
-# FIXME
+# FIXME (--as-needed)
 # Since version 10, both gcc and clang started to ignore certain linker errors
 # when compiling with -fsanitize=address. This eventually leads up to -lcrypt
 # not being correctly propagated, but the fact is masked by the aforementioned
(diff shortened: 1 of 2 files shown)


vagrant/bootstrap_scripts/arch-sanitizers-gcc.sh
```diff
@@ -49,11 +49,12 @@ rm -fr "$BUILD_DIR"
 # See:
 #   https://bugzilla.redhat.com/show_bug.cgi?id=1827338#c3
 #   https://github.com/systemd/systemd-centos-ci/issues/247
-export LDFLAGS="-Wl,--no-as-needed"
 
 meson "$BUILD_DIR" \
       --werror \
       -Dc_args='-fno-omit-frame-pointer -ftrapv' \
+      -Dcpp_args='-fno-omit-frame-pointer -ftrapv' \
+      -Db_asneeded=false `# See the FIXME (--as-needed) above` \
```
evverx, line 57: These line appears to break `scorecard`: https://github.com/ossf/scorecard/pull/1297 :-)

evverx: That's a nice trick though :-)

mrc0mmand: Hah, that's unfortunate. Should I drop it or is this fixable on the scorecard's side?

evverx: `scorecard` uses https://github.com/mvdan/sh and ideally it should be fixed there. As far as I can tell, that's exctly what was reported in https://github.com/mvdan/sh/issues/636

I'd probably keep it :-)

mrc0mmand: Ah, sweet, thanks for looking into it :)

PR #358, opened 2021-03-11
vagrant: let's run even slow tests under sanitizers

mrc0mmand:
After a bit of manual testing it looks like all tests currently tagged
as 'slow' are capable of running under sanitizers without timeouts.


keszybz:
Looks reasonable.

PR #386, opened 2021-05-25
agent/vagrant: install libbpf from repository instead of manual build

bluca:
@mrc0mmand is this enough to install the libbpf package in the image too? Context: https://github.com/systemd/systemd/pull/19705

### DIFF
--- a/agent/bootstrap.sh
+++ b/agent/bootstrap.sh
@@ -123,6 +123,7 @@ fi
     pushd libbpf/src
     LD_FLAGS="-Wl,--no-as-needed" NO_PKG_CONFIG=1 make
     make install
+    ldconfig
     popd
     rm -fr libbpf
 )
--- a/vagrant/boxes/Vagrantfile_archlinux_systemd
+++ b/vagrant/boxes/Vagrantfile_archlinux_systemd
@@ -57,6 +57,7 @@ Vagrant.configure("2") do |config|
     pushd libbpf/src
     LD_FLAGS="-Wl,--no-as-needed" NO_PKG_CONFIG=1 make
     make install
+    ldconfig
     popd
     rm -fr libbpf
 


mrc0mmand:
> @mrc0mmand is this enough to install the libbpf package in the image too? Context: [systemd/systemd#19705](https://github.com/systemd/systemd/pull/19705)

The CentOS 8 version is *ancient* (0.0.8 vs 0.4.0 from libbpf/libbpf), so I'd prefer to leave the manual install in place. However, the solution for systemd/systemd#19705 should be simple `ldconfig` call after `make install` in the libbpf step to make the newly installed libs available for `dlopen()` (or at least that's what I got from the `dlopen()` man page):

```
# make install
  INSTALL  bpf.h libbpf.h btf.h xsk.h bpf_helpers.h bpf_helper_defs.h bpf_tracing.h bpf_endian.h bpf_core_read.h libbpf_common.h
  INSTALL  libbpf.pc
  INSTALL  libbpf.a libbpf.so libbpf.so.0 libbpf.so.0.4.0
# ldconfig -p | grep bpf || echo nope
nope
# ldconfig -p | grep bpf || echo nope
	libbpf.so.0 (libc6,x86-64) => /lib64/libbpf.so.0
	libbpf.so (libc6,x86-64) => /lib64/libbpf.so
```

bluca:
Ok, updated, thanks

mrc0mmand:
> Ok, updated, thanks

Thanks! Also, could you do the same for the Arch Linux part? The libbpf there is also a bit behind (0.3.x), so we'd miss the opportunity to test the latest libbpf bits with systemd.

I'll trigger an image update right after merging this PR, so we can move forward in systemd/systemd#19705.

bluca:
Archlinux's version is... _outdated?_ What's this travesty!

PR #391, opened 2021-06-11
vagrant: run TEST-36-NUMAPOLICY under sanitizers

mrc0mmand:
This test inadvertently tests systemd-journald by using a lot of cursors
and other journald-specific stuff, stumbling across interesting issues,
like systemd/systemd#19895.


mrc0mmand:
Replaced by a test case in TEST-04-JOURNAL.

PR #463, opened 2022-02-13
vagrant(coverage): check for (possibly) missing coverage

mrc0mmand:
If the test logs contain lines like:

```
...systemd-resolved[735885]: profiling:/systemd-meson-build/src/shared/libsystemd-shared-250.a.p/base-filesystem.c.gcda:Cannot open
```

it means we're possibly missing some coverage since gcov can't write the stats,
usually due to the sandbox being too restrictive (e.g. ProtectSystem=yes,
ProtectHome=yes) or the $BUILD_DIR being inaccessible to non-root users.

### DIFF
--- a/vagrant/test_scripts/test-arch-coverage.sh
+++ b/vagrant/test_scripts/test-arch-coverage.sh
@@ -210,6 +210,26 @@ exectask "lcov_drop_gperf" "lcov -r all-integration-tests.coverage-info '*.gperf
 # in vagrant/vagrant-ci-wrapper.sh
 exectask "coveralls_upload" "coveralls --no-gcov --lcov-file integration-tests.coverage-info"
 
+# If the test logs contain lines like:
+#
+# ...systemd-resolved[735885]: profiling:/systemd-meson-build/src/shared/libsystemd-shared-250.a.p/base-filesystem.c.gcda:Cannot open
+#
+# it means we're possibly missing some coverage since gcov can't write the stats,
+# usually due to the sandbox being too restrictive (e.g. ProtectSystem=yes,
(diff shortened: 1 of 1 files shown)


evverx:
@mrc0mmand I think it's a nice way to increase the coverage but I think at some point new tests should be added :-)

By the way, I noticed that the coverage went up a little a couple days ago (without new tests). I wonder what caused it?

mrc0mmand:
> @mrc0mmand I think it's a nice way to increase the coverage but I think at some point new tests should be added :-)

Indeed, that's definitely on my radar :-)

> 
> By the way, I noticed that the coverage went up a little a couple days ago (without new tests). I wonder what caused it?

That most likely https://github.com/systemd/systemd-centos-ci/commit/57802f1eb121194b15612383801ca9b640533aad (which was prompted by https://github.com/systemd/systemd/pull/22481#issuecomment-1036081066 and also resulted in this PR).

evverx:
Thanks! What's weird though is that the badge shows 52% (instead of 58%). As far as I can see the coverage went down and then it went up but the badge somehow wasn't updated. Not that it matters though :-)

mrc0mmand:
> Thanks! What's weird though is that the badge shows 52% (instead of 58%). As far as I can see the coverage went down and then it went up but the badge somehow wasn't updated. Not that it matters though :-)

Yeah, the drop was caused by me screwing up the mentioned "fix" the first time. Let's hope the badge is just cached somewhere and gets updated soon :-)

evverx:
> Let's hope the badge is just cached somewhere and gets updated soon :-)

It seems to be the case. My browser caches that badge so I still see 52% but in incognito mode it's fine

PR #469, opened 2022-03-01
vagrant(rawhide): set SELinux to permissive before installing systemd

mrc0mmand:
otherwise we might f-up the machine due to incompatible
policies/unlabeled system files.


mrc0mmand:
The outstanding NM-related issues are known and should be resolved soon.

PR #484, opened 2022-04-11
tree-wide: significantly reduce the amount tests marked as 'flaky'

mrc0mmand:
Thanks to https://github.com/systemd/systemd/pull/22992 many of such tests should be now stable.


yuwata:
Maybe, it is better to add HOMED test to the list of the flaky tests.

PR #492, opened 2022-05-30
vagrant: try to run the networkd test suite in parallel w/ the rest

mrc0mmand:


yuwata:
Looks nice. Thank you! I guess this massively improve the performance.

mrc0mmand:
So, this seems to be working as expected, at least in the Arch runs. The only exception are the `*.dusty.ci.centos.org` nodes, where we already disable any kind of parallelism, since the machines are much slower and parallelism leads to very flaky tests (even in this case the networkd test suite failed in one tests just on the first try). So once in a while the CI job will still take the full ~2 hours, but I still think this change makes sense, since on 4 out of 5 [chassis](https://wiki.centos.org/QaWiki/PubHardware) we'll save ~30 minutes per job. 

I'll test the C8S part now with the same change, but it should, hopefully, behave the same as in the Arch case.

yuwata:
Thank you!!

PR #496, opened 2022-06-01
vagrant(upstream): run TEST-70-TPM2 under sanitizers

mrc0mmand:


evverx:
@mrc0mmand I think this particular test can be run by setting LD_PRELOAD again after it's unset in https://github.com/systemd/systemd/blob/39a742888b34c8f7c9b85a06163433122f3179aa/test/test-functions#L901 but in general I think the testsuite should always set LD_PRELOAD, "ignore" memory leaks with LSAN_OPTIONS=exitcode=0 and then look for systemd LSan backtraces in the journal: https://github.com/systemd/systemd/pull/23598. That PR is a work in progress though.

evverx:
> That PR is a work in progress though

Apparently it didn't work out. Apart from irrelevant memory leaks other issues like
```
AddressSanitizer: invalid alignment requested in aligned_alloc
```
popped up unexpectedly and I'm not sure what to do about that without complicating the script too much.

I'll try to send a PR turning on all the tests to figure out how widespread the LD_PRELOAD issue is. If it doesn't affect a lot of tests I think setting LD_PRELOAD on a case by case basis would kind of work.

evverx:
> I think setting LD_PRELOAD on a case by case basis would kind of work

Turns out it wouldn't :-) Some tests fail without LD_PRELOAD (due to binaries linked against the instrumented systemd libraries) but with LD_PRELOAD irrelevant memory leaks pop up there. Looks like the only option would be to always set LD_PRELOAD globally and unset it explicitly when bash/grep/mount are run.

evverx:
> Looks like the only option would be to always set LD_PRELOAD globally and unset it explicitly when bash/grep/mount are run.

It wouldn't help to make `TEST-25-IMPORT` pass though. There `tar` run by `systemd-importd` should be run with `LSAN_OPTIONS=exitcode=0` or `ASAN_OPTIONS=detect_leaks=0`.

mrc0mmand:
> > Looks like the only option would be to always set LD_PRELOAD globally and unset it explicitly when bash/grep/mount are run.
> 
> It wouldn't help to make `TEST-25-IMPORT` pass though. There `tar` run by `systemd-importd` should be run with `LSAN_OPTIONS=exitcode=0` or `ASAN_OPTIONS=detect_leaks=0`.

It also works with `ASAN_OPTIONS=leak_check_at_exit=0`, since the final check seems to be the one which causes all the troubles:

With `leak_check_at_exit=1`:
```strace
close(3</home/mrc0mmand/tmp/git/tar/src/test.tar.gz>) = 0
close(1</dev/pts/19>)                   = 0
close(2</dev/pts/19>)                   = 0
gettid()                                = 3398217
prctl(PR_GET_DUMPABLE)                  = 1 (SUID_DUMP_USER)
getpid()                                = 3398217
mmap(NULL, 2101248, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f9719287000
mprotect(0x7f9719287000, 4096, PROT_NONE) = 0
rt_sigprocmask(SIG_BLOCK, ~[ILL ABRT BUS FPE SEGV XCPU XFSZ], [], 8) = 0
clone(child_stack=0x7f9719487ff0, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_UNTRACED) = 3398218
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
getpid()                                = 3398217
prctl(PR_SET_PTRACER, 3398218)          = 0
futex(0x7ffe2164d9b0, FUTEX_WAKE_PRIVATE, 1) = 1
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
wait4(3398218, NULL, __WALL, NULL)      = 3398218
munmap(0x7f9719287000, 2101248)         = 0
getpid()                                = 3398217
write(2, "==3398217==LeakSanitizer has encountered a fatal error.\n", 56) = -1 EBADF (Bad file descriptor)
getpid()                                = 3398217
write(2, "==3398217==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1\n", 104) = -1 EBADF (Bad file descriptor)
getpid()                                = 3398217
write(2, "==3398217==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)\n", 77) = -1 EBADF (Bad file descriptor)
exit_group(1)                           = ?
+++ exited with 1 +++
==3398219==Attached to thread 3398214.
==3398219==Processing thread 3398214.
==3398219==Stack at 0x7ffebfdd8000-0x7ffec05d8000 (SP = 0x7ffec05d50b8).
==3398219==TLS at 0x7fb124788b00-0x7fb124789d80.
==3398219==Detached from thread 3398214.
```

With `leak_check_at_exit=0`:
```strace
newfstatat(AT_FDCWD</home/mrc0mmand/tmp/git/tar/src>, "test", {st_dev=makedev(0xfd, 0x3), st_ino=13501407, st_mode=S_IFREG|0664, st_nlink=1, st_uid=1000, st_gid=1000, st_blksize=4096, st_blocks=0, st_size=0, st_atime=1654523351 /* 2022-06-06T15:49:11.530547605+0200 */, st_atime_nsec=530547605, st_mtime=1654523351 /* 2022-06-06T15:49:11.530547605+0200 */, st_mtime_nsec=530547605, st_ctime=1654523351 /* 2022-06-06T15:49:11.530547605+0200 */, st_ctime_nsec=530547605}, AT_SYMLINK_NOFOLLOW) = 0
write(3</home/mrc0mmand/tmp/git/tar/src/test.tar.gz>, "test\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0000000664\0000001750\0000001750\00000000000000\00014247402727\000012277\0 0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0ustar  \0mrc0mmand\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0mrc0mmand\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 10240) = 10240
close(3</home/mrc0mmand/tmp/git/tar/src/test.tar.gz>) = 0
close(1</dev/pts/19>)                   = 0
close(2</dev/pts/19>)                   = 0
exit_group(0)                           = ?
+++ exited with 0 +++
```

I wonder if it would make sense to open an issue with gcc/clang or if this is an "intended" behavior.

mrc0mmand:
And in the meantime maybe add something like this to `testsuite-25.sh`:

```diff
diff --git a/test/units/testsuite-25.sh b/test/units/testsuite-25.sh
index 4119d777b3..64cca5db21 100755
--- a/test/units/testsuite-25.sh
+++ b/test/units/testsuite-25.sh
@@ -5,6 +5,16 @@ set -o pipefail
 
 export SYSTEMD_PAGER=cat
 
+if [[ -n "${ASAN_RT_PATH:-}" ]]; then
+    mv /bin/tar /bin/tar.orig
+    cat >/bin/tar <<\EOF
+#!/bin/bash
+export ASAN_OPTIONS+=":leak_check_at_exit=0"
+exec /bin/tar.orig "$@"
+EOF
+    chmod +x /bin/tar
+fi
+
 dd if=/dev/urandom of=/var/tmp/testimage.raw bs=$((1024*1024+7)) count=5
 
 # Test import
```

From a quick test it seems to be doing the right thing.

evverx:
@mrc0mmand I'll take a look a bit later today. In light of https://github.com/systemd/systemd/pull/23621 I'm trying to roll out the fuzz targets built without ASan/UBSan run against the public OSS-Fuzz corpora but `honggfuzz` seems to be crashing: https://github.com/google/honggfuzz/issues/463 :-)

evverx:
> I wonder if it would make sense to open an issue with gcc/clang or if this is an "intended" behavior.

As far as I can remember LSan is supposed to work in certain environments where there are no seccomp filters for example. Issues related to somewhat restricted environments are usually rejected and I suspect in this case LSan doesn't work because `importd` does some weird stuff somewhere to put its subprocesses (along with LSan) in a sandbox of some kind. If it isn't the case I think it should be reported.

> And in the meantime maybe add something like this to testsuite-25.sh:

I think it makes sense. In theory it should be possible to also wrap `mount`, `grep` and so on to avoid sprinkling the test scripts with `ASAN_OPTIONS`, `LD_PRELOAD` and stuff like that. I haven't tried that so I'm not sure if it's feasible though :-)

mrc0mmand:
> > I wonder if it would make sense to open an issue with gcc/clang or if this is an "intended" behavior.
> 
> As far as I can remember LSan is supposed to work in certain environments where there are no seccomp filters for example. Issues related to somewhat restricted environments are usually rejected and I suspect in this case LSan doesn't work because `importd` does some weird stuff somewhere to put its subprocesses (along with LSan) in a sandbox of some kind. If it isn't the case I think it should be reported.

I can reproduce it on my desktop (F35 ATTOW) without any restrictions:
```
$ touch test
$ LD_PRELOAD=/usr/lib64/libasan.so.6 ASAN_OPTIONS=abort_on_error=1 tar pcvf test.tar test
test
Aborted (core dumped)
```

(Also tested on Rawhide with libasan from gcc and clang with the same result.)

evverx:
Got it. I don't think `-shared-libsan` is actively maintained in `clang` so it probably should go to the gcc bug tracker (where that mode is used by default). Then again I'm not sure this use case (where uninstrumented binaries linked against uninstrumented libraries are run with `LD_PRELOAD=`) is a priority there. I think the wrapper should suffice for the time being.

mrc0mmand:
> Got it. I don't think `-shared-libsan` is actively maintained in `clang` so it probably should go to the gcc bug tracker (where that mode is used by default). Then again I'm not sure this use case (where uninstrumented binaries linked against uninstrumented libraries are run with `LD_PRELOAD=`) is a priority there. I think the wrapper should suffice for the time being.

Agreed. It looks like it's the same case with `grep` as well:

```
$ echo x >test
$ LD_PRELOAD=/usr/lib64/libasan.so.6 ASAN_OPTIONS=abort_on_error=1 grep x test
x
Aborted (core dumped)
```

```
$ LD_PRELOAD=/usr/lib64/libasan.so.6 ASAN_OPTIONS=abort_on_error=1:leak_check_at_exit=0 grep x test
x
```

so creating small wrappers for `tar` and `grep` would make sense. I'll rework https://github.com/systemd/systemd/pull/23643 in that sense.

evverx:
@mrc0mmand thanks! Now the CI should be ready for new tpm tests that hopefully should be added in PRs like https://github.com/systemd/systemd/pull/23556.

PR #497, opened 2022-06-03
vagrant(upstream): run the networkd testsuite in parallel w/ sanitizers

mrc0mmand:
Follow-up to 1eeab0fda47b21ff4931e4942a73ca06e90ca775, but for the
sanitizer run.


mrc0mmand:
Interesting, in this case the time save is quite noticeable - the time went from ~2 hours to ~1 hour and 20 minutes. I'll let this run overnight to check if it's stable and if so, we should have all CentOS CI jobs with reasonable run times once again.

PR #512, opened 2022-07-21
agent(upstream-alt): install specific kernel-tools version

mrc0mmand:
Since we're not rebooting the machine in the alt-arch runs, install
kernel-tools and other kernel packages for the currently running kernel
instead of the latest one.

### DIFF
--- a/agent/bootstrap-alt.sh
+++ b/agent/bootstrap-alt.sh
@@ -63,7 +63,9 @@ ADDITIONAL_DEPS=(
     integritysetup
     iproute-tc
     iscsi-initiator-utils
-    kernel-modules-extra
+    "kernel-modules-$(uname -r)"
+    "kernel-modules-extra-$(uname -r)"
+    "kernel-tools-$(uname -r)"
     kmod-wireguard # Kmods SIG
     knot
     libasan


mrc0mmand:
@yuwata one more issue regarding the recent kernel shenanigans - do you have any idea why `test_altname` fails here?

```
======================================================================
ERROR: test_altname (__main__.NetworkctlTests)
----------------------------------------------------------------------
  File "test/test-network/systemd-networkd-tests.py", line 901, in test_altname
    check_output('udevadm control --reload')
  File "test/test-network/systemd-networkd-tests.py", line 98, in check_output
    return subprocess.run(command, check=True, universal_newlines=text, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs).stdout.rstrip()
  File "/usr/lib64/python3.6/subprocess.py", line 438, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['udevadm', 'control', '--reload']' returned non-zero exit status 2.
```

I can't seem to find anything useful in the log, but that's probably just me.

yuwata:
I guess that's caused by the recent optimization. I guess the failure occurs randomly, not always. I will take a look.

mrc0mmand:
@yuwata unfortunately, even with https://github.com/systemd/systemd/pull/24073 `test_altname` is still not happy :(

yuwata:
Ugh, I will take a look next week.

mrc0mmand:
Let's merge this, since the change makes sense on its own and the test_altname fail is tracked in https://github.com/systemd/systemd/issues/24111.

PR #554, opened 2022-09-26
agent(upstream): move the c8s job to c9s

mrc0mmand:


mrc0mmand:
Current (potentially blocking) issues:
 * idmapped mounts are... in a weird state in RHEL9/C9S (https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/commit/c93c728360fd10a0dc6538a7fb567bc216c18c32) and would require patching upstream systemd with redhat-plumbers/systemd-rhel9@3c54c67a7fc65dc5b49b2452739c19b94eeb98a9
 * RHEL9/C9S QEMU is missing some drivers for TEST-64-UDEV-STORAGE (NVME & megasas2)

PR #581, opened 2022-12-05
agent(rhel9): build systemd EFI tools

mrc0mmand:


mrc0mmand:
The CI fail is expected (missing patch in the RHEL9 repo).

PR #584, opened 2023-01-08
vagrant(arch): build with _FORTIFY_SOURCE=3

mrc0mmand:


evverx:
@mrc0mmand could you try with `--optimization=3` just in case?

mrc0mmand:
> @mrc0mmand could you try with `--optimization=3` just in case?

Seems to be working fine as well. I wonder what combination I should leave there - personally I'd go with `_FORTIFY_SOURCE=3 -flto=auto -O1`, since `-O3` might make debugging of failed jobs slightly harder (and since LTO has been a source of headaches in the past, I'd leave it there as well).

evverx:
I think it should probably just resemble Fedora builds as closely as possible. Actual buffer overflows should be caught by the ASan job with full backtraces and so on so they should be more or less debuggable. "Buffer overflows" popping up in the _FORTIFY_SOURCE job will most likely be caused by new versions of gcc or something like that. Having said that I'm not sure if it should be run on every PR.

evverx:
Just to clarify I think ideally it should probably be run on every PR to test systemd builds shipped by actual distributions but it can just clog the Cent OS CI with no benefits in terms of testing. I'm not sure what to do about the build test either. Ideally the unit tests should be run there but GHActions are already clogged and even a daily cron job could prevent some PRs from being tested in a timely manner.

mrc0mmand:
> I think it should probably just resemble Fedora builds as closely as possible. Actual buffer overflows should be caught by the ASan job with full backtraces and so on so they should be more or less debuggable. "Buffer overflows" popping up in the _FORTIFY_SOURCE job will most likely be caused by new versions of gcc or something like that. Having said that I'm not sure if it should be run on every PR.

Completely agree. I'll try to leverage Fedora infrastructure for this, since there an option to use the Testing Farm, but I have a blocker opened for systemd which prevents us from using it. Let me try to push this a bit.

Issue #614, opened 2023-04-11
missing debug symbol in upstream CentOS CI ??


yuwata:
Per-test coredumpctl output shows many 'n/a'. See [TEST-25-IMPORT_1_coredumpctl_collect_FAIL.log](https://jenkins-systemd.apps.ocp.cloud.ci.centos.org/job/upstream-centos8/1306/artifact/systemd-centos-ci/artifacts_unufi4gi/testsuite-logs-upstream.owz/TEST-25-IMPORT_1_coredumpctl_collect_FAIL.log).

But in [coredumpctl_collect_FAIL.log](https://jenkins-systemd.apps.ocp.cloud.ci.centos.org/job/upstream-centos8/1306/artifact/systemd-centos-ci/artifacts_unufi4gi/testsuite-logs-upstream.owz/coredumpctl_collect_FAIL.log), all symbols are resolved.

PR #616, opened 2023-04-11
Set $STRIP_BINARIES _before_ creating the base image

mrc0mmand:
Otherwise it has no effect in tests that use the base image.

Resolves: #614


evverx:
FWIW I think `STRIP_BINARIES` should be set to `no` by default in `test-functions`. As far as I can remember it came from dracut (where it makes sense) but the test suite shouldn't be stripped.

(This comment isn't in any way prompted by valgrind refusing to work with stripped binaries :-)).

mrc0mmand:
> FWIW I think `STRIP_BINARIES` should be set to `no` by default in `test-functions`. As far as I can remember it came from dracut (where it makes sense) but the test suite shouldn't be stripped.
> 
> (This comment isn't in any way prompted by valgrind refusing to work with stripped binaries :-)).

Yeah, definitely agree. I'll post a follow-up PR to the systemd repo to see what's going to break :)

mrc0mmand:
-> https://github.com/systemd/systemd/pull/27215

PR #622, opened 2023-06-03
agent(alt): use unified cgroups in nspawn as well

mrc0mmand:
Let's try this one more time.


mrc0mmand:
The TEST-74 fail should be fixed by https://github.com/systemd/systemd/pull/27907.

PR #623, opened 2023-06-04
vagrant: bump to v2.3.7

mrc0mmand:


mrc0mmand:
Installation of v2.3.6 is borked on C8S, but nightly of v2.3.7 works, so let's wait a bit.

PR #630, opened 2023-06-21
tree-wide(upstream): install stress

mrc0mmand:
For https://github.com/systemd/systemd/pull/28064#issuecomment-1599945252


anitazha:
Thanks! Didn't realize it was so easy to add utilities for the extended tests

PR #673, opened 2023-11-08
vagrant(arch): prefix .network file with digits

yuwata:
To make the file has higher precedence over the default .network files.

For https://github.com/systemd/systemd/pull/29930.


mrc0mmand:
Thank you! I killed the job and adjusted some knobs behind the scenes, so it automagically propagates to the "production" images. Once the queue frees up, it should start again.

Issue #678, opened 2023-11-14
upstream: please set -Ddefault-network=true 


yuwata:
Currently, it disabled by default. So, no CIs currently test the feature. Please enable the option.

PR #679, opened 2023-11-14
vagrant(arch): install default .network files in the Arch job

mrc0mmand:
Since the option is disabled by default, let's enable it in the Arch Linux job to make sure we don't break it in the future.

Resolves: #678


yuwata:
Thank you!!

PR #701, opened 2024-01-31
Revert "agent(rhel9): pin kernel to kernel-5.14.0-407.el9.x86_64"

mrc0mmand:
RHEL-22465 has been resolved and a fixed kernel is already on the mirrors (5.14.0-412.el9.x86_64), so let's revert the kernel pin.

This reverts commit 63dbbd8b44b37557854d6faec8b0e8b9235f1450.


mrc0mmand:
(The fixed version is -414, I just can't read apparently, so this will have to wait a bit.)

PR #723, opened 2024-04-08
vagrant(rawhide): explicitly pull libkmod into the initrd

mrc0mmand:
With [0] merged we need to pull in libkmod explicitly; drop this once [1] lands in Fedora.

[0] https://github.com/systemd/systemd/pull/31131$
[1] https://github.com/dracut-ng/dracut-ng/pull/118


mrc0mmand:
Just merged https://github.com/systemd/systemd-centos-ci/pull/724 that apart from installing the latest systemd and selinux-policy also installs dracut-ng from git (the CI fails there due to AVCs, but that's unrelated to the dracut change).

PR #734, opened 2024-10-29
agent(upstream): pin tzdata to 2024a-2.el9.noarch

yuwata:
Workaround for https://github.com/systemd/systemd/issues/34471


mrc0mmand:
Thank you! Please add this to https://github.com/systemd/systemd-centos-ci/blob/main/agent/bootstrap.sh instead, since `bootstrap-rhel9.sh` is just for downstream RHEL9 runs. That should, hopefully, do the trick.

yuwata:
Done. PTAL.

mrc0mmand:
Thank you!

